PR #85, opened 2021-07-12
Feature/1.x theming default search/76

### LINKED ISSUE #76: Theme re-factor - Default Search
# Sample: http://localgov.dev.mark.ie/search

[] Put the templates in place in the theme
[] Move styling into localgov_base theme
[] confirm feature parity

markconroy:
Closes #76


msayoung:
@markconroy 
I've tested locally, as your site seems to have moved on (looking good btw)

Could the Apply button display inline? 

I believe on the other theme we decided (though I don't think it was implemented) to show the button as a magnifying glass with the Apply text hidden with the `.visually-hidden` class. Happy to leave this though as it's probably best to be consistent with the other searches on the site. 

The order of the items seems to have changed from the original which had: 

1. H1
2. Search box
3. H2 - Search results
4. Showing items... 
5. Results

image
image

Could we have the search results view mode title as an H3 please? 

The original search results didn't have visible numbers on the search results, though this doesn't bother me particularly. I don't think we are showing them on other search results, so I guess we should be consistent. 

Not sure if this is a config issue or a theme one - some types of content (eg directory venue) don't seem to have their view mode set up correctly. eg search for collaborators

msayoung:
sorry commented in the wrong place - moved https://github.com/localgovdrupal/localgov_base/issues/76

PR #91, opened 2021-07-15
Fix for region display issue in localgov_base.theme

andrew-wallis:


markconroy:
Hi @andrew-wallis 

Thanks for this. I've added it to my local version, but it doesn't seem to be working. 

Here's my step-by-step page when using your code (you can see the sidebar is missing):



Here it is with the current code we have in the repo (you can see there is a sidebar):

Screenshot 2021-07-15 at 16 10 26

andrew-wallis:
Thanks for taking a look @markconroy. I missed that there was an additional has_sidebars variable in there. I've added that in and is working on my local build of the site.

markconroy:
Brilliant work @andrew-wallis 

Merged now. I'll create a new release straightaway.

andrew-wallis:
Great @markconroy ! Was nice to contribute something back.

Issue #130, opened 2021-09-01
Safari outlines the main content


drinkynet:
Safari (iOS and Mac) outlines the main element when its clicked or tapped on as [it has a tabindex](https://github.com/localgovdrupal/localgov_base/blob/1.x/templates/layout/page.html.twig#L103). It may need outline:none applying to it to hide it.

![iOS safari outline on main element](https://user-images.githubusercontent.com/6780886/131735287-e4bd7bf4-e6c4-4dd4-9db6-e609e7fd095b.png)
![MacOS safari outline on main element](https://user-images.githubusercontent.com/6780886/131735482-36ee9161-477e-448a-8dc2-21ee19b35044.png)

finnlewis:
@markconroy noticed in our OxDUG meetup!

markconroy:
Here's a PR for this issue 
https://github.com/localgovdrupal/localgov_base/pull/131

danchamp:
Nice catch, thanks @drinkynet

markconroy:
Closed by #131

PR #131, opened 2021-09-07
sets main-content id outside of main element

markconroy:
PR for #130

### DIFF
--- a/templates/layout/page.html.twig
+++ b/templates/layout/page.html.twig
@@ -100,7 +100,9 @@
   {{ page.messages }}
 {% endif %}
 
-<main id="main-content" tabindex="-1"> {# The "skip to content" link jumps to here. #}
+<a id="main-content" tabindex="-1"></a> {# The "skip to content" link jumps to here. #}
+
+<main class="main" id="main"> 
   {% if has_content_top %}
     {{ page.content_top }}
   {% endif %}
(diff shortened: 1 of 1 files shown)


danchamp:
@markconroy I think we want to keep the id on the main element, since it's the landmark that we want to be the target of the skip link.

The issue is the focus outline that Safari displays, so it may be a better approach to remove that in CSS?

Also paging @msayoung for a sense-check.

markconroy:
@danchamp 

Does it need to be the landmark? Drupal core does not link to a landmark, it links to a `a` tag, exactly like I have done. You can see it in the bartik theme's `/templates/page.html.twig`

danchamp:
The screen reader experience in Safari seems to be much better with the main element as the target. On VoiceOver, the target gets announced as:

> Main, you are currently on a heading level one.

With the empty anchor element:

> You are currently on a link, inside web content

GovUK and NHS use the main element as the target, but interestingly don't have the tabindex set for the element.

Going to dig further...

Hmm, I understood that without the tabindex there was an issue with the element reliably receiving focus, but that doesn't appear to be the case. GovUK removed it some time ago:

https://github.com/alphagov/govuk_elements/pull/534

And have found this article which confirms there's no longer a need for that negative tabindex:

https://knowbility.org/blog/2019/skip-links

So, I think we can fix this by just removing `tabindex="-1"` from `main`, and leaving it as the target of the skip link.

markconroy:
Cool stuff. I'll get that fixed later so. Thanks Dan.

markconroy:
Hi @danchamp 

This is ready again now. Thanks.

Issue #135, opened 2021-09-10
Subsite page banner image floats left


finnlewis:
Hi @markconroy 

On this subsite page: https://demo.localgovdrupal.org/bereavement-services/choosing-memorial

The banner image could do with central alignment of some sort. 

![image](https://user-images.githubusercontent.com/326588/132862376-5ba16a6a-13b3-4222-8119-f8ecb20241f6.png)

markconroy:
Thanks @finnlewis 

I noticed that last week, but forgot to create an issue for it. I think that banner is supposed to go full width, rather than just centered. I'll get that done next week.

PR #137, opened 2021-09-16
adds correct class for banner__image

markconroy:
PR for #135


finnlewis (approved):
This works for me locally - images now full width like on the overview.

finnlewis:
Thanks @markconroy !

PR #151, opened 2021-10-12
Default theming for forms

markconroy:

PR #152, opened 2021-10-15
PR for Release 1.2.0 - default forms theming

markconroy:
This is the PR that adds default theming for forms. Any changes, we can create further issues.


andybroomfield (approved):
As this is about releasing PR #151 that is already approved and no other changes, this should be released.

PR #159, opened 2021-11-17
PR for 1.2.2

markconroy:
Included:
#158


markconroy:
Merged, as there is only 1 PR in this so no point in getting it verified a second time.

PR #161, opened 2022-01-06
Allow outline in maps to be set via CSS variable

### LINKED ISSUE #160: Directory areas should pick up colours from base theme pallet
When a directory item has an area, the outline of the area is a light blue colour, which is not in use anywhere else on the site.

Can the styling pick up a colour and tint from an existing colour that's already been defined in the theme

**Example URL**
[demo.localgovdrupal.org/national-parks-uk](https://demo.localgovdrupal.org/national-parks-uk?f%5B0%5D=localgov_directories_facets%3A8)

**Example screenshot**
image

markconroy:
This PR fixes #160 

We now have a variable set in the base theme to allow us to override the default blue outline we have for maps. 

The variable is called `--leaflet-map-outline-stroke-colour`. It's currently set to the same colour blue as the default map, so current sites will not see any difference unless they set this variable in the variables.css file.

### DIFF
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -381,4 +381,7 @@ body {
   --alert-banner-inner-margin-horizontal: 0;
   --alert-banner-inner-padding-vertical: var(--spacing);
   --alert-banner-inner-padding-horizontal: 0;
+
+  /* Leaflet Mapping */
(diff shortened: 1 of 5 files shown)


ekes (approved):

andybroomfield (approved):
This works (able to change it in sub theme)

Issue #171, opened 2022-01-24
The generated sub theme name is still LGD_SUB_THEME_NAME


andybroomfield:
When running the create subtheme script, the theme name as called in THEMENAME.info.yml still uses the name
`LGD_SUB_THEME_NAME`.
This should be updated to use the theme name that was entered when running the generation script.
<img width="1401" alt="Screenshot 2022-01-24 at 2 14 09 pm" src="https://user-images.githubusercontent.com/1467480/150799376-f8aa2808-dca2-4e34-a3a0-182d98683693.png">

andybroomfield:
I'm also getting this error
> User warning: The following theme is missing from the file system: LGD_SUB_THEME in Drupal\Core\Extension\ExtensionPathResolver->getPathname() (line 63 of core/lib/Drupal/Core/Extension/ExtensionPathResolver.php).

andybroomfield:
This is on MacOS 10.14 (I'll check on a more recent one later).
PHP 7.4

Also tried with Lando (using lando ssh and running the bash command).
This works ok.

PR #173, opened 2022-01-24
fixes theme name not getting rewritten on mac after subtheme script

markconroy:
This seems to fix the issue, using `perl` instead of `sed` for issue #171


andybroomfield (approved):
This works great, thank you.

PR #188, opened 2022-04-22
adds relative position for main canvas so toolbar is not obstructed

### LINKED ISSUE #187: Set main canvas div to position:relative
In future (e.g. when creating an off-canvas region for microsites, or perhaps for some hero positioning) we might need to use `position: absolute;` to place some items.

This will cause these  items to cover the toolbar when they are active. Setting the main canvas region to relative will stop this.

markconroy:
Fixes #187

### DIFF
--- a/css/layout/layout-utilities.css
+++ b/css/layout/layout-utilities.css
@@ -2,6 +2,14 @@
   @file Utility classes to drop into templates.
 */
 
+/* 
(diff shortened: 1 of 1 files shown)


finnlewis (approved):
Discussing this with @markconroy , it won't break any existing sites but is useful for sub-themes.

PR #194, opened 2022-05-04
Release 1.2.9

stephen-cox:
* Updated GitHub workflow

* Coding standards fixes

### DIFF
--- a/localgov_base.libraries.yml
+++ b/localgov_base.libraries.yml
@@ -90,7 +90,7 @@ page-title-block:
       css/components/page-title-block.css: {}
 
 service-landing-page:
-  css:) ? TRUE : FALSE;
+  css:
     theme:
       css/components/service-landing-page.ie11.css: {}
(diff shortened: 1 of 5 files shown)


localgov_base.libraries.yml
```diff
@@ -90,7 +90,7 @@ page-title-block:
       css/components/page-title-block.css: {}
 
 service-landing-page:
-  css:
+  css:) ? TRUE : FALSE;
```
finnlewis: @stephen-cox is this line meant to be like this?

stephen-cox: @finnlewis Bloody IDE pasting stuff it shouldn't. Fixed now

finnlewis (approved):
Nice one!

localgov_base.libraries.yml
finnlewis: That's what the second pair of eyes are for!

PR #200, opened 2022-05-11
Service Page Border Variables

3mkay:
Additional variable for one of the service page borders.

### DIFF
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -390,4 +390,7 @@ body {
 
   /* Leaflet Mapping */
   --leaflet-map-outline-stroke-colour: #38f;
+  
+  /* Service Landing Page */
+  --service-cta-border: var(--border-width);
 }
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -394,5 +394,5 @@ body {
   --leaflet-map-outline-stroke-colour: #38f;
   
   /* Service Landing Page */
-  --service-cta-border: var(--border-width);
+  --service-cta-border: var(--border);
 }
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -98,7 +98,7 @@ body {
   --font-size-h5: var(--font-size-medium);
   --font-size-h6: var(--font-size-medium);
 
-  /* Borders */
+  /* Border elements */
   --border-width: 1px;
   --border-width-large: 5px;
   --border-color: var(--color-grey-light);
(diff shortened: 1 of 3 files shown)


css/base/variables.css
```diff
@@ -390,4 +390,7 @@ body {
 
   /* Leaflet Mapping */
   --leaflet-map-outline-stroke-colour: #38f;
+  
+  /* Service Landing Page */
+  --service-cta-border: var(--border-width);
```
markconroy: I tihnk this should evaluate to `--border`, shouldn't it?

css/components/service-cta-block.css
```diff
@@ -1,7 +1,7 @@
 .service-cta-block__list {
   padding-left: 0;
   list-style: none;
-  border-bottom: var(--border);
+  border-bottom: var(--service-cta-border);
```
markconroy, line 4: This is fine, but 'only' if `--service-cta-border` evaluates to `--border` (and then it also won't break existing sites.

css/components/teaser.css
```diff
@@ -55,7 +55,7 @@
 .news-article .field--name-localgov-news-related > .field__item {
   margin-bottom: var(--vertical-rhythm-spacing);
   padding-bottom: var(--vertical-rhythm-spacing);
-  border-bottom: var(--border);
+  border-bottom: var(--service-cta-border);
```
markconroy, line 58: If we are going to set the border on teasers to something specific, let's called it `--border-teaser` or `--teaser-border`, and let that evaluate to `--border` in the `variables.css` file

markconroy (changes requested):

css/base/variables.css
3mkay: Indeed. I missed the declaration of that in amongst the others. 

Any issue if I separate those meta variables out into their own section to make them more obvious?

markconroy: Yes, work away, happy to look at a PR for that to see how it works.

3mkay: It's only a small tweak, have included it in this PR [9ef8f6c](https://github.com/localgovdrupal/localgov_base/pull/200/commits/9ef8f6cd9b68aee15e11963eff159d97d82ed10d)

markconroy:
@3mkay 

Looks like we have a few conflicts here since this PR was created. Can you look at those (I reckon they are quite small) and we'll get this merged asap then. Thanks.

PR #202, opened 2022-05-17
Add and apply new css variables to override heading styles

msayoung:


css/base/variables.css
```diff
@@ -98,6 +98,32 @@ body {
   --font-size-h5: var(--font-size-medium);
   --font-size-h6: var(--font-size-medium);
 
+  /* Heading colours */
+  --text-colour-dark: #000000;
```
markconroy: I'm a little concerned about setting this as a direct colour rather than to another variable that already exists.

By setting it to equal the variable that it should have already been equating too, we should not break existing sites.

Something like this
`--text-colour-dark: var(--color-black);`

msayoung: Ah yes, I was waiting to see what we were doing with the --text-colour-dark and --text-colour-light field on the microsites, but I'll change this for now and we can change again if we need to.

PR #207, opened 2022-05-24
Set the video to 100% width

msayoung:


markconroy:
@msayoung I'm merging this because the CSS is perfect for what we need, but I think we should have a follow up in localgov_subsites for the video field formatter to out put a larger video. Currently it only outputs a 200px wide video.

msayoung:
> 

@markconroy 
I did have a look at that, but as they currently don't have a video template even I decided it made more sense here, where we do have one. But I totally agree :)

markconroy:
Not having a template should not matter. The field formatter should still out put the video at a higher quality than 200px. I'm worried that what you have done is exactly right, but we'll be stretching 200px wide videos to 100%, where we should be containing 768px wide videos _down_ to 100% for better video quality.

PR #220, opened 2022-06-14
Add a template for the new link component

msayoung:


finnlewis:
@markconroy another one that @msayoung would like your eyes on before merging. :)

css/components/link.css
```diff
@@ -0,0 +1,14 @@
+.link-block__title-wrapper {
+  display: flex;
+  align-items: center;
+  margin-bottom: var(--vertical-rhythm-spacing);
+}
+
+.link-block__title-icon {
+  flex-shrink: 0;
+  margin-right: var(--spacing-smaller);
```
markconroy: We should probably create variables for these margins, so it's easier for sub-themes to override them

css/components/link.ie11.css
```diff
@@ -0,0 +1,7 @@
+.link-block__title-wrapper {
+  margin-bottom: 1.5rem;
+}
+
+.link-block__title-icon {
+  margin-right: 0.5rem;
```
markconroy: Let's use variables for these margins

templates/paragraphs/paragraph--localgov-link.html copy.twig
```diff
@@ -0,0 +1,80 @@
...
+
+{% set title_icon = 'chevron-right' %}
+
+<div{{ attributes.addClass(classes) }}>
+
+  {% if paragraph.localgov_url %}
+    <div class="link-block__title-wrapper">
+      {% include "@localgov_base/includes/icons/icon.html.twig" with {
+          icon_name: title_icon,
+          icon_wrapper_element: 'div',
```
markconroy: I _think_ `div` is the "sensible default", so does not need to be included here as it is set automatically.

templates/paragraphs/paragraph--localgov-link.html copy.twig
```diff
@@ -0,0 +1,80 @@
+{#
```
markconroy: Looks like this file is a duplicate

css/components/link.css
msayoung: I realised we didn't need that margin-bottom anyway

templates/paragraphs/paragraph--localgov-link.html copy.twig
msayoung: do you mean the sensible default for the icon? 
if so then this div is wrapping the icon and the text and flexing them. but I may have misunderstood :)

templates/paragraphs/paragraph--localgov-link.html copy.twig
msayoung: gone

templates/paragraphs/paragraph--localgov-link.html copy.twig
markconroy: I mean the div on line 64 `icon_wrapper_element: 'div',` - the icon template uses `div` by default, so you only need to put a value in there is you want to use `span` or something elsse.

css/components/link.css
markconroy: Should we create a variable for `margin-right` here? Something like `--link-title-icon-margin`?

PR #226, opened 2022-06-29
Add video library to the libraries

msayoung:


markconroy (commented):
Is this video library used anywhere?

msayoung:
It _should_ be being used here https://github.com/localgovdrupal/localgov_base/blob/1.x/templates/paragraphs/paragraph--localgov-video.html.twig but we forgot to define it

localgov_base.libraries.yml
```diff
@@ -273,3 +273,9 @@ leaflet:
     theme:
       css/components/leaflet.ie11.css: {}
       css/components/leaflet.css: {}
+
+video:
+  css:
+    theme:
+      css/components/video.ie11.css: {}
```
markconroy: Let's remove this ie11 line. We don't have any ie11 specific video css (I don't think).

After that, merge merge merge!

msayoung: done. I'm hitting the merge button.
I guess we'll need a release at some point?

Issue #234, opened 2022-07-06
Block won't appear


rupertj:
I've had an issue today where a block I've created would appear with the title displayed, but hid when the title was hidden, or would display just fine alongside another block in the same region, regardless of if the title was set to display or not.

I've tracked this down to the code in localgov_base_preprocess_page() which renders the regions and strips out HTML tags to determine if the region is empty or not. Currently it strips out all tags, which means that blocks whose content is entirely HTML will not display.

I've fixed this by adding exceptions to the stip_tags call for input and form, which lets my block display as expected.

markconroy:
This is fixed from PR #235

PR #235, opened 2022-07-06
Stopped form and input tags being stripped out when determining if a region is empty or not.

rupertj:
Fixes issue #234

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -59,7 +59,7 @@ function localgov_base_preprocess_page(&$variables) {
     else {
       $rendered = Drupal::service('renderer')->renderPlain($variables['page'][$region]);
     }
-    $variables[$has] = !empty(trim(strip_tags($rendered, '<form><input>')));
+    $variables[$has] = !empty(trim(strip_tags($rendered, '<drupal-render-placeholder><embed><hr><iframe><img><input><link><object><script><source><style><video>')));
   }
   $variables['has_sidebars'] = $variables['has_sidebar_first'] || $variables['has_sidebar_second'];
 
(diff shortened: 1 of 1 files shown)


markconroy:
HI @rupertj 

Does your form have no content at all? As in, is there not even text on a submit button or inside a label that would cause our check to validate?

rupertj:
It's an exposed form from views being used as a search form, so there's a placeholder in a single field, and then a button to submit it.

Here's the markup if you're curious:

```
<div class="views-exposed-form block block-lbhf-search block-lbhf-sitewide-search-block" data-drupal-selector="views-exposed-form-search-default" id="block-lbhfsitewidesearchblock">
    <form action="/search" method="get" id="views-exposed-form-search-default-block" accept-charset="UTF-8">
        <div class="js-form-item form-item js-form-type-search-api-autocomplete form-item-s js-form-item-s form-no-label">
            <input data-drupal-selector="edit-s" placeholder="Search" type="text" id="edit-s" name="s" value="" size="30" maxlength="128" class="form-text required" required="required" aria-required="true">
        </div>
        <div data-drupal-selector="edit-actions" class="form-actions js-form-wrapper form-wrapper" id="edit-actions">
            <input data-drupal-selector="edit-submit-search" type="submit" id="edit-submit-search" value="Apply" class="button js-form-submit form-submit">
        </div>
    </form>
</div>
```

Having thought about it a bit more, there's other scenarios that could also fail, like a linked image in a block.

rupertj:
I was just reading the d.o issue linked in the code, and found the last comment interesting: https://www.drupal.org/project/drupal/issues/953034#comment-14588869 . The contrib module linked there uses this set of tags as ones that should remain when determining if a region is empty or not:
```
[
    '<drupal-render-placeholder>',
    '<embed>',
    '<hr>',
    '<iframe>',
    '<img>',
    '<input>',
    '<link>',
    '<object>',
    '<script>',
    '<source>',
    '<style>',
    '<video>',
];
```

markconroy:
Looks good. Thanks @rupertj

PR #242, opened 2022-07-13
Apply radio styles to 'Radios other' elements

### LINKED ISSUE #241: Radio buttons with Webform element type 'Radios other' are not being styled
Webform offers two types of element for groups of radio buttons: 'Radios', and 'Radios other'.

They have different wrapper classes, and localgov_base currently only targets the former, and styles are not being applied to radios within 'Radios other' elements.

danchamp:
Closes #241 

Extends form styling to include radio buttons in Webform 'Radios other' elements.

Because the 'Radios other' wrapper will also include a text input, some rules need to be more specific than default radio styling.

### DIFF
--- a/css/components/form-items.css
+++ b/css/components/form-items.css
@@ -200,7 +200,8 @@ input[type="submit"]:hover {
   wrapping them.
 */
 
-.webform-type-radios .form-item {
+.webform-type-radios .form-item,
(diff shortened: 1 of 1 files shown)


cjstevens78 (approved):
looks good to me - thanks

PR #255, opened 2022-07-26
sets input fields to use the primary font

### LINKED ISSUE #254: Use theme font in forms
Currently the form items - input, textarea, select, etc - are using the user agent stylesheet for fonts. Let's explicitly set them to use the same font as the body of the website `--font-primary`.

markconroy:
Closes #254


finnlewis (approved):
Looks good to me @markconroy but let me know if you would like me to manually test it.

PR #262, opened 2022-08-03
stops columns in guide nav overspilling to next column

### LINKED ISSUE #261: Don't break items in guide nav onto next column
When some items in our guide nav have long titles, they can spill over into the next column, which is quite jarring.

Take for example the last item in the second column here which overflows into a first item in the third column.
Screenshot 2022-08-03 at 14 49 07

markconroy:
Closes #261

Before:

Screenshot 2022-08-03 at 14 49 07

After:
Screenshot 2022-08-03 at 14 54 12

### DIFF
--- a/css/components/guide-nav.css
+++ b/css/components/guide-nav.css
@@ -17,6 +17,7 @@
 .lgd-guide-nav__list-item {
   margin-bottom: var(--spacing);
   font-size: var(--font-size-large);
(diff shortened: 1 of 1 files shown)


seeduardo:
Looks good 👍

finnlewis (approved):
Not tested locally, but looks good to me too and a trivial enough change!

PR #263, opened 2022-08-08
Change background colour on guide page navigation

keelanfh:
The background colour on guide pages isn't WCAG2.1AA compliant with `--color-link` (`--color-accent`) used for links in this block (both in base and Scarfolk, as well as other colours people might commonly choose):

image

I've suggested a fix for this where we set the background colour to `--color-grey-lighter`, which should give better contrast with a range of colours.

Incidentally, `--color-grey-light` is actually darker than `--color-grey-medium` - there might be a good reason for this, but it does seem a bit counterintuitive when you look down the list of CSS variables...

Same change suggested for `.service-landing-page__contact` for the same reason - the contact box includes email and web addresses which get link styling applied too.

### DIFF
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -49,7 +49,7 @@ body {
   --color-white: #fff;
   --color-grey-lightest: #f9f9f9;
   --color-grey-lighter: #eaeaea;
-  --color-grey-light: #96918b;
+  --color-grey-light: #cecfd0;
   --color-grey-medium: #b1b4b6;
   --color-grey-dark: #505a5f;
   --color-black: #0b0c0c;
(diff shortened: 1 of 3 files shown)


finnlewis:
Thanks @keelanfh - looks pragmatic to me. 

Perhaps @markconroy could you check and merge if all good?

markconroy:
I noticed this myself last week. 

I think we should change --color-grey-light to a colour that is lighter than --color-grey-medium for sure.

We fixed this in the `scarfollk` theme (kind of - at least for the sidebar) by setting the sidebar background colour to --color-grey-lightest instead of --color-grey-light, but let's fix it properly here now instead.

markconroy:
@keelanfh  do you want to update the PR to _also_ set --color-grey-light to something lighter than the default grey?

keelanfh:
@markconroy I've updated that now - from a search, it doesn't look like that variable is used anywhere else, so shouldn't have any implications.

PR #265, opened 2022-08-09
adds sitewide search theming

markconroy:
Looks like we are not attaching our sitewide search library to our theme, and even if we were, we could still make things look a lot nicer by default.

## Before:

Screenshot 2022-08-09 at 12 55 23

## After:

Screenshot 2022-08-09 at 12 58 19


finnlewis (approved):
Nice one @markconroy - definitely an improvement!

PR #271, opened 2022-08-15
Added form element template

MattOz-CDS:
Added a form element template in order to facilitate validation styling as per GDS design pattern recommendations

### DIFF
--- a/templates/form/form-element.html.twig
+++ b/templates/form/form-element.html.twig
@@ -1,0 +1,94 @@
+{#
+/**
+ * @file
+ * Theme override for a form element.
+ *
+ * Available variables:
+ * - attributes: HTML attributes for the containing element.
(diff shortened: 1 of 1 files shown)


markconroy:
Hi @MattOz-CDS 

What exactly is this doing that different to our current form-element.html.twig? And do we need any CSS/JS to go with it to style the GDS-style validation?

MattOz-CDS:
Hi @markconroy,  thanks for looking at this. I couldn't see a form-element template in the current 1.x branch. The purpose of this form-element.html.twig is to display the form error messages above the input field.

markconroy:
Hi @MattOz-CDS 

Thanks for this. I'll merge it in now, but I think we're going to need to have a follow-up created to add some CSS for form errors.

PR #273, opened 2022-08-23
The File title shouldn't be output as a link

### LINKED ISSUE #272: Outputting file title as a link causes error for anonymous users
User who don't have access to the file just uploaded are shown a string.
The current code means this causes an error as it assumes a link is to be rendered.

MR incoming

alanburke:


finnlewis:
@markconroy are you happy enough with this? Tests pass and looks OK to us.

markconroy:
Thanks Finn. I'll merge it so (we are using it on one of our LGD sites with no issues).

PR #281, opened 2022-09-21
Add blog post template to localgov_base

msayoung:


markconroy (approved):

markconroy (approved):

templates/content/node--localgov-blog-channel--full.html.twig
```diff
@@ -0,0 +1,126 @@
...
+        {{ metadata }}
+      </div>
+    </footer>
+  {% endif %}
+
+  <div{{ content_attributes.addClass('blog-channel__content', 'node__content') }}>
+
+    {#{{ content|without('localgov_news_search', 'localgov_news_facets', 'localgov_blog-channel_all_view') }}#}
+    <div class="lgd-row">
+      <div class="lgd-blog--featured">
```
markconroy: If this is inside a `.lgd-row` it will need to have a class of `.lgd-row__full` (or whatever width we want it to be.

templates/content/node--localgov-blog-channel--full.html.twig
```diff
@@ -0,0 +1,126 @@
...
+      <div{{ author_attributes.addClass('node__submitted') }}>
+        {% trans %}Submitted by {{ author_name }} on {{ date }}{% endtrans %}
+        {{ metadata }}
+      </div>
+    </footer>
+  {% endif %}
+
+  <div{{ content_attributes.addClass('blog-channel__content', 'node__content') }}>
+
+    {#{{ content|without('localgov_news_search', 'localgov_news_facets', 'localgov_blog-channel_all_view') }}#}
```
markconroy: Let's remove this commented out line

templates/content/node--localgov-blog-channel--full.html.twig
markconroy: Let's only have a `.lgd-blog--featured` if it is at the same level as a `lgd-blog` class. Might be better to change this to `blog-channel__featured` instead.

templates/content/node--localgov-blog-post--teaser.html.twig
```diff
@@ -0,0 +1,138 @@
+{#
```
markconroy, line 1: Is there any chance this template could use the `node--teaser.html.twig` template instead of being its own template?

templates/content/node--localgov-blog-post--teaser.html.twig
```diff
@@ -0,0 +1,138 @@
...
+  {{ attach_library('localgov_base/teaser') }}
+  {{ attach_library('localgov_base/blog') }}
+{% endif %}
+
+<article{{ attributes.addClass(classes).removeAttribute('role') }}>
+
+  <div{{ content_attributes.addClass('lgd-teaser__content', 'lgd-teaser__content--' ~ content_type|clean_class) }}>
+
+    {{ title_prefix }}
+    {% if label and not page %}
```
markconroy: This conditional is probably not needed. We always have a `label` since it's a required field. And we'll never have `page` since it's a not-full-view-mode

templates/content/node--localgov-blog-post--teaser.html.twig
```diff
@@ -0,0 +1,138 @@
...
+          {{ content.localgov_blog_categories[key] }}
+          {# Add a comma between values, but not the last item #}
+          {% if not loop.last %}{{',' | t}}{% endif %}
+        {% endfor %}
+      {% endif %}
+
+      {# To do - hide this when we are on the blog channel #}
+      {{'-' | t}}{{content.localgov_blog_channel}}
+    </div>
+  </div>
```
markconroy: I think if we use the `node--teaser.html.twig` as our base, we can put this in here as a blog

```
{% block blog_meta %}
  all this html
{% endblock %}
```

templates/content/taxonomy-term--localgov-blog-author.html.twig
```diff
@@ -0,0 +1,47 @@
...
+ * @see template_preprocess_taxonomy_term()
+ */
+#}
+{% if not localgov_base_remove_css %}
+  {{ attach_library('localgov_base/term') }}
+  {#{{ attach_library('localgov_base/blog') }}#}
+{% endif %}
+
+<div{{ attributes }}>
+  <div class="lgd-term--content">
```
markconroy: We should only have `lgd-term--content` if we have it at the same level as `lgd-term`, it's a modifier of the block `lgd-term`, so something like `lgd-term lgd-term--content` or if it's an element, `lgd-term__content` (making sure we have `lgd-term` as an ancestor.

templates/content/taxonomy-term--localgov-blog-author.html.twig
```diff
@@ -0,0 +1,47 @@
...
+ *   - id: The ID of the taxonomy term.
+ *   - bundle: Machine name of the current vocabulary.
+ * - view_mode: View mode, e.g. 'full', 'teaser', etc.
+ *
+ * @see template_preprocess_taxonomy_term()
+ */
+#}
+{% if not localgov_base_remove_css %}
+  {{ attach_library('localgov_base/term') }}
+  {#{{ attach_library('localgov_base/blog') }}#}
```
markconroy: If this commented out code is not being used, let's delete it.

templates/content/taxonomy-term--localgov-blog-author.html.twig
```diff
@@ -0,0 +1,47 @@
...
+{% endif %}
+
+<div{{ attributes }}>
+  <div class="lgd-term--content">
+    {% if term.localgov_blog_author_image.value %}
+      <div class="lgd-term--content__image">
```
markconroy: We can't have elements of modifiers in BEM CSS. This class will need to be something like `lgd-term__image`.

markconroy (changes requested):

templates/content/taxonomy-term--localgov-blog-author.html.twig
```diff
@@ -25,13 +25,12 @@
 #}
 {% if not localgov_base_remove_css %}
   {{ attach_library('localgov_base/term') }}
-  {#{{ attach_library('localgov_base/blog') }}#}
 {% endif %}
 
 <div{{ attributes }}>
```
markconroy: We'll need an 'lgd-term' class in here: `<div{{ attributes.addClass('lgd-term') }}>`

templates/content/taxonomy-term--localgov-blog-author.html.twig
```diff
@@ -25,13 +25,12 @@
...
-  <div class="lgd-term--content">
+  <div class="lgd-term">
```
markconroy: I think we can remove this, if we add the class to the parent (or leave it here, and we won't bother adding it to the parent/attributes)

templates/content/taxonomy-term--localgov-blog-author.html.twig
```diff
@@ -25,13 +25,12 @@
...
     {% if term.localgov_blog_author_image.value %}
-      <div class="lgd-term--content__image">
+      <div class="lgd-term--image">
```
markconroy: We need to make these classes elements instead of modifiers: `lgd-term__image`

css/components/term.css
```diff
@@ -1,21 +1,21 @@
-.lgd-term--content__image img {
+.lgd-term--image img {
```
markconroy: We need to fix up the CSS classes here to reflect the BEM classes that we will use in the HTML. `lgd-term__image` instead of `lgd-term--image`, etc.

markconroy (changes requested):

markconroy (approved):

PR #284, opened 2022-09-26
Add colour for the H2 in this section, as was causing a false-postive…

msayoung:
… with a11y checker

### DIFF
--- a/css/components/region-tabs.css
+++ b/css/components/region-tabs.css
@@ -9,6 +9,10 @@
   background-color: black;
 }
 
+.lgd-region--tabs h2 {
+  background-color: white;
+}
+
 .lgd-region--tabs .tabs > ul {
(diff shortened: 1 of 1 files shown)


css/components/region-tabs.css
```diff
@@ -9,6 +9,10 @@
   background-color: black;
 }
 
+.lgd-region--tabs h2 {
```
markconroy, line 12: Let's not do this. It means every `h2` that ever gets into the tabs region for whatever reason will have a white background. And that might lead us to having to have lots of overrides (potentially).

Let's set this class as `.lgd-region--tabs .visually-hidden { background-color: var(--color-white);}`

markconroy (changes requested):

markconroy (approved):

PR #293, opened 2022-10-07
sets media files to link to their file

### LINKED ISSUE #292: Media files are no longer linking to the file
Looks like #273 means that any media documents output as "generic file" or "table of files" are not linking to the files.

markconroy:
Closes #292

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -88,16 +88,18 @@ function localgov_base_preprocess_file_link(&$variables) {
 
   // 123.45 KB -> 123.45KB
   $variables['file_size'] = strtr($variables['file_size'], [' ' => '']);
-
+  
   if (isset($variables['link']['#title'])) {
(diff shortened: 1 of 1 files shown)


ekes (approved):
So this shouldn't really be necessary, because at link render array is passed around https://git.drupalcode.org/project/drupal/-/blob/608eabd8e1f318caec8b18bddcfbea0d874a0145/core/modules/file/file.module#L1532 When you upload media to the private file system as anonymous within core media for example a link does get produced, you get an access denied if you try and follow it though.

Webform seems to break this contract https://git.drupalcode.org/project/webform/-/blob/e00b05d92397acd3437bc535b2c2a99a502ed61b/includes/webform.theme.inc#L671 by providing just a string... even though it's still a link template and you can expect that render array :-( 

Switching based on `[#type]` might be even clearer why this is being done, but `[#title]` works.

PR #296, opened 2022-10-07
If the map has a value is_empty and it's true don't display.

### LINKED ISSUE #229: (Empty) map displays in directory channel
Steps to reproduce: 

1. add a directory channel
2. add promo pages
3. visit channel 
4. see an empty map

ekes:
The value is set for views without any location features in localgov_directories_leaflet_map_view_style_alter().

PR https://github.com/localgovdrupal/localgov_directories/pull/239
Fixes https://github.com/localgovdrupal/localgov_directories/issues/229#issuecomment-1271418506

### DIFF
--- a/templates/misc/leaflet-map.html.twig
+++ b/templates/misc/leaflet-map.html.twig
@@ -18,6 +18,7 @@
  */
 #}
 
-{{ attach_library('localgov_base/leaflet') }}
-
-<div id="{{ map_id }}" style="min-width: 150px; {% if height is not empty %}height: {{ height }}{% endif %}"></div>
+{% if not map.is_empty %}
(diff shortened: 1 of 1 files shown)


markconroy (commented):
This PR looks fine, but surely if the map is_empty, then we should never get as far as this template at all, or should we?

msayoung:
> This PR looks fine, but surely if the map is_empty, then we should never get as far as this template at all, or should we?

@markconroy  see ekes' Argh here: https://github.com/localgovdrupal/localgov_directories/issues/229#issuecomment-1271418506

markconroy:
In that case, this PR looks good to me.

markconroy (approved):

finnlewis:
You want to merge and release @markconroy ?

PR #311, opened 2022-10-24
sets colour for title on death of person alert

### LINKED ISSUE #310: Title colour for death of a notable person alert banner
Currently the title is black on a black background.

markconroy:
Closes #310

### DIFF
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -416,7 +416,10 @@ body {
   --alert-banner-bg-color-major: var(--color-danger);
   --alert-banner-bg-color-notable-person: var(--color-black);
   --alert-banner-color: var(--color-white);
-  --alert-banner-title-color-notable-person: var(--color-white);
+  --alert-banner-title-color-notable-person: inherit;
+  --alert-banner-title-color-announcement: inherit;
+  --alert-banner-title-color-minor: inherit;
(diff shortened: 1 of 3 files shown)


andybroomfield:
I'd be happy with this, however see [my comment](https://github.com/localgovdrupal/localgov_alert_banner/pull/225#discussion_r1003475205) as I think there is a better solution.

````css
.localgov-alert-banner .localgov-alert-banner__title {
  color: inherit;
}
````

markconroy:
@andybroomfield `inherit` is fine, but means you need to add specific CSS to override it in your sub-theme. I think that's appropriate for the CSS in your module, but in the base theme we are trying to make things override-able via a CSS variable. So in this case, to change the color of that title, you just need to reset what that variable is in your `variables.css` file.

andybroomfield:
@markconroy fair enough, just an FYI, the bug exists for all the alert banner kinds, not just notable person. Its just we don't notice it since we set the colour of the text to black on those banners. The h2 colour is currently setting it so they will change if someone changes the h2 colour.

Screenshot 2022-10-25 at 9 29 58 am

markconroy:
PR updated, so we can cover each banner type.

andybroomfield (approved):
This works great now, thanks @markconroy

PR #318, opened 2022-11-24
Add some default styles for the embed views paragraph

msayoung:


finnlewis:
@Adnan-cds hopes to be able to look at this one, as they use localgov_base ! :)

Adnan-cds:
Sorry, I am struggling to test this on a stock LocalGov install.  I found out later we don't have this *Embedded content list* Paragraph type in our site, so decided to try it on an empty LocalGov site.  I was trying to add a Page component of this type and it wouldn't let me, saying "There is no display available. Please select another view or change the field settings.":
screenshot-embed-views-error
Have I missed any thing?

msayoung:
Thanks for trying @Adnan-cds 
I think we only get predefined views in the microsite ecosystem, which makes me think this is pretty safe to merge. 
I'll fix the coding standards though

Adnan-cds:
> which makes me think this is pretty safe to merge.

Okay, I will leave this with Finn given my lack of understanding of Microsites at the moment.  If Finn approves it then I am happy for it to go in.

PR #325, opened 2022-12-01
Randomise IDs on the submit button with the ID edit-actions

msayoung:


markconroy (approved):

Adnan-cds:
Just wanted to say that this has broken the [LocalGov Address lookup](https://github.com/localgovdrupal/localgov_forms/blob/1.x/src/Plugin/WebformElement/LocalgovWebformUKAddress.php) Webform element :(  I am yet to figure out the cause.

msayoung:
sorry @Adnan-cds - shall we roll it back?

Adnan-cds:
Nothing to be sorry about :)  Let me try to figure out if the problem is in localgov_forms or in localgov_base_preprocess_container().

Adnan-cds:
Hmm... tricky one.  And will remain so until this has been [fixed in core](https://www.drupal.org/project/drupal/issues/1852090).

The following change at least allows the address lookup element to work.  But it is possible there are other forms which may still be affected.  Which is why I am not offering this as a blanket solution.  Could you please check if it retains the accessibility improvements?

```
diff --git a/localgov_base.theme b/localgov_base.theme
index 2e46107..4616a26 100644
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -123,10 +123,10 @@ function localgov_base_preprocess_container(&$variables) {
     $id = $variables['element']['#id'];
     if ($id == 'edit-actions') {
       $id .= '--' . Crypt::randomBytesBase64(8);
+      $variables['attributes']['id'] = $id;
+      $variables['attributes']['data-drupal-selector'] = $id;
+      $variables['element']['#attributes']['data-drupal-selector'] = $id;
+      $variables['element']['#id'] = $id;
     }
-    $variables['attributes']['id'] = $id;
-    $variables['attributes']['data-drupal-selector'] = $id;
-    $variables['element']['#attributes']['data-drupal-selector'] = $id;
-    $variables['element']['#id'] = $id;
   }
 }
```

PR #330, opened 2022-12-14
2022 w50 remove extraneous tabs from twig files

ctorgalson:
Minor coding standards fix: replaces all tab characters in *.html.twig files with two spaces.

(I can't find specific references to tabs in Drupal Twig or Twig coding standards, but all one or all the other is better than a mix IMO).

@markconroy: here you go.


markconroy (approved):
Looks good to me, thanks @ctorgalson

markconroy:
@msayoung Do you want to give this a once-over?

msayoung (approved):

Issue #332, opened 2022-12-16
header.js does not ensure operations happen only once; unnecessary reassignment of context


ctorgalson:
Drupal behaviors should run only when the DOM elements they apply to are included in the `context` parameter passed to the function. On the _initial_ load, context will be equivalent to `document`. On subsequent runs (on e.g. AJAX requests etc), `context` will contain DOM elements that've changed.

The contents of context are used to determine whether or not to perform whatever operations the behavior is responsible for _only_ as many times as is necessary. We usually do this by exiting when no relevant DOM elements are found, and/or by 'marking' the DOM elements affected by the behaviour with a class or data attribute. This can be done manually or with [`once()`](https://www.npmjs.com/package/@drupal/once).

`header.js` misuses `context` by populating it with the contents of `document` even when it's empty, meaning we may discover DOM elements in it that have _not_ been updated. In addition, `header.js` doesn't correctly ensure that it runs only once or only when necessary.

This issue discusses both of these because of how they're related.

## Reassignment of `context`

Currently, `header.js` duplicates [core's assignment of the context variable](https://git.drupalcode.org/project/drupal/-/blob/9.5.x/core/misc/drupal.js#L18) on [line 14](https://github.com/localgovdrupal/localgov_base/blob/1.x/js/header.js#L14):

```
      context = context || document;
```

### Fixing this

This can just be removed.

## header.js does not ensure operations happen only once

There are several checks in the file to prevent operations from happening unnecessarily (see lines [49-51](https://github.com/localgovdrupal/localgov_base/blob/1.x/js/header.js#L49-L51), [54-55](https://github.com/localgovdrupal/localgov_base/blob/1.x/js/header.js#L54-L55), [172-173](https://github.com/localgovdrupal/localgov_base/blob/1.x/js/header.js#L172-L173), but it's still doing certain things multiple times.

This isn't _necessarily_ harmful but it can impede performance, and repeated processing can be the source of weird bugs e.g. when an event handler gets added multiple times to the same element. And as it happens, `header.js` is doing exactly that):

![image](https://user-images.githubusercontent.com/1063424/208116976-fe4b7f42-fbce-4e6b-ac40-519584e9334d.png)

As the screenshot shows, the time isn't a major issue here, but I _have_ been seeing the primary toggle misbehave in a way consistent with having multiple identical event listeners, but I have _not_ confirmed this issue is responsible.

### Fixing this

[...]

PR #334, opened 2023-01-02
332 header js does not ensure operations happen only once unnecessary reassignment of context

ctorgalson:
This PR's main intention is to address the issues raised in #332, but I've used that as an excuse for streamlining and regularizing the code--but without substantially altering the logic.

Reproducing the commit message from f78af4e, here's a list of the basic changes:

  - trailing whitespace trimmed,
  - whitespace in general attended to,
  - remove a LOT of unnecessary querySelector() calls,
  - consolidate most-used DOM references in navInfo variable,
  - regularize ternaries (there were three or more styles in use),
  - eliminate several element.classList .contains() / .add() / .remove()
    ternaries via the use of element.classList.toggle(),
  - collect variables declarations/assignments at the beginning of the file,
  - separate the header search label manipulation from the nav/button work,
  - use variables instead of repeating class names and selectors,

As far as I can see, I haven't broken anything :grimacing: 

Incidentally, for reviewers, I didn't mention this specifically in commit messages, but I've changed all instances of this structure:

```js
      let foo;
      let bar;
      if (context.querySelector('.foo')) {
        foo = context.querySelector('.foo');
      }
      if (context.querySelector('.bar')) {
        bar = context.querySelector('.bar');
      }
```

...to this structure:

```js
  const foo = context.querySelector('.foo');
  const bar = context.querySelector('.bar');
```

...because:

  - it's shorter,
  - these variables are never redefined (i.e. `const`, not `let`),
  - it's more efficient (half the `querySelector()` calls),
  - the result if the DOM element is missing is similar (now the result will be ~empty `NodeList` objects~ `null` instead of `undefined` ~variables~)

### DIFF
--- a/js/header.js
+++ b/js/header.js
@@ -68,23 +68,23 @@ if (window.NodeList && !NodeList.prototype.forEach) {
       // Which menu region to show is decided by the "toggleThatWasClicked" parameter.
       function handleToggleClick(toggleThatWasClicked) {
         const currentState = toggleThatWasClicked.getAttribute('aria-expanded');
-        currentState === 'false' ? 
-          toggleThatWasClicked.setAttribute('aria-expanded', 'true') : 
+        currentState === 'false' ?
+          toggleThatWasClicked.setAttribute('aria-expanded', 'true') :
           toggleThatWasClicked.setAttribute('aria-expanded', 'false');
-          
-        toggleThatWasClicked.classList.contains('lgd-header__toggle--active') ? 
+
+        toggleThatWasClicked.classList.contains('lgd-header__toggle--active') ?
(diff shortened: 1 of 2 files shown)


finnlewis:
Leaving this one for @markconroy and @msayoung to test.

ekes:
Just logging my comments: once makes complete sense; and code looks much nicer. I'm not sure if there are automated tests for this? So I think it certainly needs one other person doing and install and manually confirming no issues.

finnlewis:
Thanks @ekes 

Let's let @markconroy  and / or @msayoung test this and approve.

markconroy:
This looks good to me, tested it on the site I am currently working on.

PR #336, opened 2023-01-05
335 Fix focus state text wrapping issue

### LINKED ISSUE #335: Focus indicator does not appear correctly on <a> elements wrapping on multiple lines in Chrome, Edge
Examples:

image
image
image

I think this is a WCAG 2.1 AA fail because without the underline, the focus indicator does not have sufficient contrast against white backgrounds.

Correct behaviour (working in Safari):

image

keelanfh:
To fix #335 

Just tested in the browser inspector (I copied this change from GOV.UK and it appears to work), I didn't build and test a site. Would be great if someone could test in Chrome and Edge as those are the browsers where I observed this issue.

### DIFF
--- a/css/base/base.css
+++ b/css/base/base.css
@@ -58,6 +58,7 @@ a:focus {
   background-color: var(--color-focus);
   -webkit-box-shadow: 0 -2px var(--color-focus), 0 4px var(--color-black);
   box-shadow: 0 -2px var(--color-focus), 0 4px var(--color-black);
+  -webkit-box-decoration-break: clone;
 }
 
 h1,


keelanfh:
On further investigation, this issue came up due to a change in Chromium 108.

See issue on GOV.UK Frontend which describes this and the change they made: https://github.com/alphagov/govuk-frontend/pull/3107

Chrome requires the vendor-prefixed `-webkit-box-decoration-break` rather than `box-decoration-break` - I assume this is added later by the GOV.UK Frontend tech stack

keelanfh:
Applied the patch to a fresh install, and it looks OK with the demo content.

finnlewis:
Thanks @keelanfh !

We can review at Merge Monday meeting at 2pm today, unless anyone has a change to take a look sooner.

finnlewis (approved):
@markconroy approves this!

PR #346, opened 2023-01-12
allows icons to be called from any theme

### LINKED ISSUE #345: Allow the icon include to use icons from outside of localgov_base
Currently you can only include icons in your theme using  `{% include icon thing with icon_name = "facebook" %}` for icons that are already in the localgov_base repo.

Let's make this more generic so other themes can easily use it.

markconroy:
Closes #345


finnlewis:
Testing locally

finnlewis (approved):
Works as expected locally, thanks @markconroy !

PR #347, opened 2023-01-16
Use the stable9 as base theme rather than the deprecated stable theme

### LINKED ISSUE #344: Stable theme has been removed from core
As described here: https://www.drupal.org/node/3309392. Stable is deprecated in Drupal 9 and removed in Drupal 10 (https://www.drupal.org/node/3309392)

The new stable base theme is called `stable9`. The alternatives are to used the `starterkit_theme` or copy everything required from the `stable` base theme and not have a base theme.

stephen-cox:
Fixes #344


finnlewis:
Hey @markconroy - does this look sound to you?

markconroy:
This is a little trickier than just setting stable9 as the base, for the basic reason that stable9 is not enabled by default.

For this PR to work, will we need an update hook somewhere to enable stable9. I've updated the PR branch to add the update_hook to `localgov_base.install`. Any takers for testing it (make sure to clear your cache for the new `.install` file to be picked up)?

finnlewis:
I'll test now

finnlewis:
Initially this is not working for me. 

 - Fresh install of localgov profile
 - Enable localgov_demo module
 - Switch to this branch
 - clear the cache 
 - drush updb (no updates)
 - error:

`146   20/Jan 15:57   php             Error      Drupal\Core\Theme\MissingThemeDependencyException: Base theme stable9 has not been installed. in Drupal\Core\Theme\ThemeInitialization->getActiveThemeByName() (line 115 of /var/www/html/we  `

finnlewis:
Why would the update hook not be firing?

markconroy:
Maybe we need to clear the cache more than once?

This was a tricky one for me over the Christmas when upgrading mark.ie to D10. Especially with stable9 not being in the list of themes on the appearance page, so you can't manually go an enable it.

finnlewis:
I can't get this hook_update_N to fire from the theme.

If I move the hook to a module, it works OK and enables the stable9 theme as expected. 

Does the update hook work for you @markconroy ?

finnlewis:
This one is still puzzling. Any thoughts @markconroy ?

finnlewis:
This documentation is a bit confusing and might be out of date: https://www.drupal.org/docs/8/themes/bs-base/theme-development/theme-update-hooks

markconroy:
> Does the update hook work for you @markconroy ?

Yep, it worked for me. But maybe it was because I changed to this branch and _then_ wrote the update hook and _then_ cleared cache again.

I'll have to test it again with the update hook already in the .install file.

markconroy:
I can't get it to fire on a clean install either. If I try running it from a module it looks like it will work, but not sure that's good enough for us, unless we add the update hook to localgov_core or something?

```
╰─ fin cr && fin drush updb
 [success] Cache rebuild complete.
 --------------- ----------- --------------- ------------------------------------ 
  Module          Update ID   Type            Description                         
 --------------- ----------- --------------- ------------------------------------ 
  localgov_demo   9502        hook_update_n   9502 - Implements hook_update_N().  
 --------------- ----------- --------------- ------------------------------------ 


 Do you wish to run the specified pending updates? (yes/no) [yes]:
 > n


In UpdateDBCommands.php line 67:
              
  Cancelled.  
```

finnlewis:
@markconroy , can we add it to the localgov profile?

stephen-cox:
I can't get the update hook to work either. I propose adding the hook to the profile and checking the base theme is enabled before enabling stable9.

Anyone using the base theme outside of LocalGov would need to manually enable stable9, so I suggest making this a minor release and ensuring this is added to the release notes.

stephen-cox:
I've removed the install file and added the hook to enable stable9 here localgovdrupal/localgov_base#347

markconroy:
@stephen-cox  maybe don't bother checking if localgov_base is installed and just go install stable9 across the board. Without that it might be hard for anyone creating their own base theme to work with it as it can't be installed via any UI, so nice for us to have the "dirty work" done for everyone.

finnlewis (approved):
This now works for me without error, when testing with https://github.com/localgovdrupal/localgov/tree/fix/2.x/localgob_base-347-enable-stable9 / https://github.com/localgovdrupal/localgov/pull/492

stephen-cox:
Discussed at Merge Monday and the preference is to only enable stable9 when needed and so to keep the condition.

Issue #352, opened 2023-01-25
In header.js primary toggle now only toggles primary nav


ctorgalson:
This is a regression introduced in #332, #334. Mea culpa.

[Line 87 of `header.js`](https://github.com/localgovdrupal/localgov_base/blob/1.x/js/header.js#L87) is currently:

```js
navInfo.primary.region.classList.toggle(regionActiveClass);                                                                                       
```
To match [the original functionality](https://github.com/localgovdrupal/localgov_base/blob/1.2.28/js/header.js#L96-L100), it _should be_:

```js
navInfo.primary.region.classList.toggle(regionActiveClass);                                                                                       
navInfo.secondary.region.classList.toggle(regionActiveClass);
```
PR follows shortly.

finnlewis:
Thanks @ctorgalson , what actually breaks and what does the PR fix?

Just wondering how urgent a release is.

ctorgalson:
Sorry for that @finnlewis :(

In the code we cleaned-up, clicking or tapping `button.gd-header__toggle--primary` would toggle the state of _every_ `.lgd-header__nav`.

In other words, clicking that button would toggle the primary _and_ secondary nav open or closed. I haven't looked at tons of LGD sites, but I can see that e.g. https://www.lbbd.gov.uk/ relies on this behaviour.

The default way this works (and a way to check an individual site) is:

- site at a breakpoint < 768px
- primary nav has content
- secondary nav has content
- primary and secondary nav are hidden until the primary nav toggle is clicked
- when the primary nav toggle is clicked, _both_ navs become visible

The PR restores this behaviour.

finnlewis:
Cool, thanks @ctorgalson - I will aim to test and release quickly to reduce the chances of anyone hitting this.

PR #353, opened 2023-01-25
Makes sure primary toggle toggles both nav regions

ctorgalson:
This PR should restore the functionality broken by #334.

- Re: 352-in-header-js-primary-toggle-now-only-toggles-primary-nav


finnlewis (approved):
Confirmed the behaviour and that this pull request fixes it.

See https://github.com/localgovdrupal/localgov_base/issues/352#issuecomment-1403839000 for description.

Current behaviour of the menu toggle with localgov_demo: 

image

With this pull request: 

image

Let's get this merged and released, thanks @ctorgalson !

ctorgalson:
Don't thank me @finnlewis, it was me that broke it in the first place :facepalm:.

Issue #367, opened 2023-02-08
HTML5 form validation 


gavin-webstuff:
HTML 5 validation tooltip is not displayed on forms on chrome 

**Steps to reproduce** 

- any form seems to demonstrate the bug so long as the upon validation there is a need to scroll to element triggered

![html5_no_validation_message](https://user-images.githubusercontent.com/1315627/217530948-a33acf15-0507-402c-b447-51ca7c66e9c8.jpg)
![html5_validation_message](https://user-images.githubusercontent.com/1315627/217530986-36336d85-a91f-42b0-87b5-6c679bd31c03.jpg)

andybroomfield:
Although this is focused on user login, and we should still fix the issue anyway for these cases, we should move away from using HTML5 validation if we want to match GDS guidelines.
see https://design-system.service.gov.uk/patterns/validation/

markconroy:
I've created a follow up issue for disabling HTML5 validation: https://github.com/localgovdrupal/localgov/issues/503

markconroy:
I'm going to close this issue since we have the related issue now in the localgovdrupal/localgov repo.

PR #368, opened 2023-02-08
fix for HTML5 validation messages

gavin-webstuff:
addresses an issue where HTML5 validation messages where not being displayed 

see issue https://github.com/localgovdrupal/localgov_base/issues/367

### DIFF
--- a/css/base/base.css
+++ b/css/base/base.css
@@ -7,6 +7,9 @@
 
 html {
   box-sizing: border-box;
+}
+
+body {
   scroll-behavior: smooth;
 }
 


markconroy (changes requested):
Thanks for this PR @gavin-webstuff. Just a few small comments:

1. We already have a `body` section in `base.css`. Let's move this change to that section (lines 26 - 32 of base.css)
2. Lines 20 to 24 of base.css handles the `scroll-behaviour` that is currently attached to the `html` element for people who `prefers-reduced-motion`. Let's move that block to _after_ the body element, and then update it to target the `body` element instead of the `html` element.

PR #370, opened 2023-02-13
Fix html5 validation tooltip

gavin-webstuff:


gavin-webstuff:
following up from https://github.com/localgovdrupal/localgov_base/pull/368

finnlewis:
Thanks @gavin-webstuff

PR #379, opened 2023-02-17
updates menu toggle to use label for show/hide, and sets text as aria-hidden

### LINKED ISSUE #377: Ensure visible label and accessible name match for hamburger menu
### Impact

Users of voice recognition software will speak the visible label in order to interact with elements – for example ‘Click Menu’. However, if the underlying accessible name does not match the command will not be recognised 

### WCAG Fail 

This item is a Fail of the WCAG 2.1 AA guidelines

### Expected results

User of voice recognition software speaks ‘Click menu’ and the hamburger menu is activated and menu content expands

### Suggested fix

Remove the aria-label ('toggle primary navigation') from the <button> element 

### Steps to reproduce

Load page, reduce viewport until mobile menu is shown, open developer tools and inspect the element. Notice the aria-label

image

markconroy:
Closes #377 

I've taken a different approach than that suggested in #377, which was to remove the aria-label. Instead, I'm setting the text to 'aria-hidden' so it's ignored by screenreaders (in case we ever decide to put in an image/icon/whatever, we can't be sure it will always be text). Instead, then, we use JS to update the aria-label from 'Open primary navigation' to 'Close primary navigation'.


msayoung:
I don't think that will solve the issue I'm afraid. 

Firstly - we don't need the Open / Close prefixes. I think that confuses the issue.

The problem is, as I understand it, is that a person who is sighted and using a voice recognition software needs to be able to guess at the accessible name of that button. 

Where there is a visible label, then we should use this. Ideally there would always be a visible label for this reason. However, when there isn't a visible label we need to use the most commonly used word. 
While "Menu" is inaccurate (don't get me started) it is commonly used . 

So my suggestion is to remove the aria-label from the button, which will default the label to the text within it. If people replace this with an icon they'll need to add alternative text / aria-label to the icon.

markconroy:
That's updated again @msayoung if you want to give it a check.

msayoung:
Hey @markconroy - tweaked that , also removed the aria-label on the Services menu, which didn't really need it either. 

Who approves this now?

msayoung (approved):

PR #380, opened 2023-02-17
sets a border colour for minor alert banner close buttons

### LINKED ISSUE #375: "Hide" button on alert banners does not meet colour contrast requirement in hover state
... I think?

image

markconroy:
Closes #375


keelanfh (approved):
Tested and looks good to me in Chrome!

msayoung (approved):

andybroomfield (approved):
Also works for me. I have opened an issue on https://github.com/localgovdrupal/localgov_alert_banner/issues/238 so we should do the same or do the suggested solid black in the modules css.

PR #382, opened 2023-02-22
Null check for secondary menu

siliconmeadow:
We found that if we weren't using the secondary menu, this would throw an exception.


markconroy (approved):

markconroy:
Thanks very much @siliconmeadow

siliconmeadow:
Thanks @markconroy - it was a joint effort with @nbrettell and me.

PR #389, opened 2023-02-27
sets correct class for secondary menu columns

### LINKED ISSUE #387: Services menu is single-column since removal of IE11 support in 1.4.0
secondary-menu.ie11.css set the column-count for any element with class `menu` within the secondary region:

https://github.com/localgovdrupal/localgov_base/blob/32261e5237ceb20012de41e1f4fd103efa1aace1/css/components/secondary-menu.ie11.css#L25

secondary-menu.css sets the column-count in a rule that targets a class that doesn't appear to exist in the LGD codebase:

https://github.com/localgovdrupal/localgov_base/blob/bd8b480f86587bc804bdafceb2faf91a91e75b61/css/components/secondary-menu.css#L30

Verify by setting an earlier version of localgov_base as default theme, and inspecting the service menu. column-count is being derived from the IE11 CSS file:

Screenshot 2023-02-23 at 13 50 53

markconroy:
Closes #387

### DIFF
--- a/css/components/secondary-menu.css
+++ b/css/components/secondary-menu.css
@@ -27,7 +27,7 @@
     background-color: var(--secondary-menu-bg-color-second);
   }
 
-  .menu--services-list {
+  .region-secondary-menu .menu {
     overflow-y: initial;
     max-height: auto;
     column-count: var(--secondary-menu-column-count);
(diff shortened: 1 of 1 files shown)


danchamp:
@markconroy One other use of that class at line 16 I think.

markconroy:
@danchamp  updated.

danchamp (approved):
Thanks @markconroy, looks good to me.

msayoung (approved):

PR #392, opened 2023-03-04
Remove our focus state from toolbar items

### LINKED ISSUE #391: Low contrast in "Manage" and user profile buttons in focus state
Screenshot 2023-03-04 at 08 36 18

In the focused state, the text on the "Manage" and user profile button doesn't meet colour contrast requirements. This is because our focus states are being partially applied here - the `background-color` is set, but the `color` is not.

Admin theme (Gin?) has:
```css
.toolbar .toolbar-bar .toolbar-item {
    color: #fff;
}
```

but 
```css
.toolbar .toolbar-bar {
    color: #ddd;
    background-color: #0f0f0f;
    box-shadow: -1px 0 3px 1px rgba(0,0,0,0.3333);
}
```

so on `a.toolbar-item` elements, the `a:focus` `background-color` takes precedence over Gin's defaults (as Gin's defaults are inherited from a parent), but `a:focus` `color` has no effect because there's a more specific CSS rule in place. This combination means we end up with white text on yellow.

We've seen similar issues crop up in three other places too at Essex, in our own theme - it's fairly common to set `background-color` on a parent element and then just `color` on the specific element, which results in the same problem. I don't know that there's anything much to be done in general, but it's just something to be aware of.

PR on the way to fix this specific issue.

keelanfh:
Removes the effect of our focus state from admin toolbar items, to resolve the color contrast issue in #391.

The default focus states are not amazing (and I'm not sure they'd meet the new requirements in WCAG 2.2) but I think they're WCAG 2.1 AA compliant, and definitely better than the white-on-yellow text we have now.

An alternative approach would be to apply our focus states correctly, however, this would require some modifications to the icons on these menus so that they meet colour contrast requirements in the focus state too.

Before:

Screenshot 2023-03-04 at 08 36 18

After:

Screenshot 2023-03-04 at 09 12 26

### DIFF
--- a/css/base/base.css
+++ b/css/base/base.css
@@ -51,14 +51,13 @@ a:hover {
   text-decoration-thickness: max(3px, 0.1875rem, 0.12em);
 }
 
-a:focus {
+a:not(.toolbar a):focus {
   text-decoration: none;
   color: var(--color-black);
   outline: 3px solid transparent;
   background-color: var(--color-focus);
-  -webkit-box-shadow: 0 -2px var(--color-focus), 0 4px var(--color-black);
   box-shadow: 0 -2px var(--color-focus), 0 4px var(--color-black);
-  -webkit-box-decoration-break: clone;
+  box-decoration-break: clone;
 }
 
 h1,
--- a/css/components/admin-toolbar.css
+++ b/css/components/admin-toolbar.css
@@ -1,3 +1,8 @@
 .toolbar-menu {
   font-size: 0.8125rem;
 }
+
+.toolbar .toolbar-bar .toolbar-item:focus {
+    background-color: transparent;
+    box-shadow: none;
+}


finnlewis:
In Merge Monday , not front enders here. Leaving this for @markconroy to review

finnlewis:
This looks good to me. 
Worth a quick check with @markconroy or @msayoung  if they are around.

msayoung (approved):

markconroy:
Merged (after a few more small changes I made)

keelanfh:
This causes a regression of #336 - even up-to-date versions Chrome and Edge require the vendor prefix on `box-decoration-break`.

image

keelanfh:
This causes a regression of #336 - even up-to-date versions Chrome and Edge require the vendor prefix on `box-decoration-break`.

image

markconroy:
According to MDN Chrome and Edge should not need that: https://developer.mozilla.org/en-US/docs/Web/CSS/box-decoration-break

But you are correct, it's needed. I've this merged now (and added back in for the step-by-step buttons). I'll create a new release straight away

keelanfh:
MDN says it's required - the `-x-` underneath signifies that, and if you click on the Chrome cell, it does say that. It's not the clearest design though!

Thanks :)

markconroy:
Well spotted. Thanks Keelan.

PR #408, opened 2023-03-16
sets all links to inline-block by default

### LINKED ISSUE #396: Newsroom teasers don't seem to show correct focus state
Would expect this to show the same focus state as other links

image

markconroy:
Closes #396


keelanfh (approved):
LGTM - has resolved this issue, and from a cursory look at a few other pages, everything looks fine.

PR #413, opened 2023-03-20
sets variable for call out box title

### LINKED ISSUE #410: Colour contrast issue on call out box title
image

The default styling from the base theme has a black header on the purple. Can we change this to white please :)

markconroy:
Closes #410

### DIFF
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -359,8 +359,8 @@ body {
   --call-out-box-padding: var(--spacing-large);
   --call-out-box-link-padding: var(--spacing-smaller);
   --call-out-box-bg-color: var(--color-accent);
-  --call-out-box-title-color: var(--color-white);
   --call-out-box-text-color: var(--color-white);
+  --call-out-box-title-color: var(--call-out-box-text-color);
 
   /* Service Statuses */
   --service-statuses-padding: var(--spacing);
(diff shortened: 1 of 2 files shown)


finnlewis:
@keelanfh @Adnan-cds @andybroomfield could you check if this would have any unintended or unexpected effect on your sites? 

I think we should be able to release this with details in release notes. 

@msayoung has some more ideas: change the  --call-out-box-title-color: to the colour of the body text for the callout box.

msayoung (approved):

keelanfh:
No effect on us - we use white text in call-out boxes anyway...

markconroy:
We've had no objections to this after a week, so merging now.

Adnan-cds:
Too late, but noticed that the fix doesn't cover h5 and h6 tags even though the [heading level field](https://github.com/localgovdrupal/localgov_paragraphs/blob/f532cf3511d5d76ea161f2e84faf0f7d74d672bf/modules/localgov_subsites_paragraphs/config/install/field.storage.paragraph.localgov_heading_level.yml#L23) provides these heading options along with h1-h4.

Also, wondering if [localgov_colour_theme](https://github.com/localgovdrupal/localgov_paragraphs/blob/2.x/modules/localgov_subsites_paragraphs/config/install/field.field.paragraph.localgov_call_out_box.localgov_colour_theme.yml) field is related to this in anyway.

markconroy:
@Adnan-cds I left h5 and h6 out of the list as I presumed we'd never have a call out box so deep in the hierarchy. But we can easily add it in "just in case".

PR #418, opened 2023-03-22
adds visually hidden h2 to announce service list

### LINKED ISSUE #417: Skipped heading level in service landing pages
Our service landing page has a h1, then a listing of teasers which all have h3 for their headings, so we are skipping h2.

Let's put a visually-hidden h2 in place to announce "Services List" or something like that just before the teaser list.

markconroy:
Closes #417

### DIFF
--- a/templates/content/node--localgov-services-landing--full.html.twig
+++ b/templates/content/node--localgov-services-landing--full.html.twig
@@ -48,6 +48,7 @@
       <div class="lgd-row__full">
         {# Begin Services List #}
         {% if node.localgov_destinations.value %}
+          <h2 class="visually-hidden">{{ 'Services List'|t }}</h2>
           <section class="service-landing-page__services lgd-teaser-list">
             {{ content.localgov_destinations }}
           </section>


msayoung:
@markconroy - have seen this and am pondering Keelan's approach in #417  and the discussion https://github.com/localgovdrupal/localgov_services/issues/188

finnlewis:
@msayoung did you have any more thoughts on this? 

@keelanfh would this still work with your approach?

keelanfh:
Given that we've now created our own template for these pages, this change wouldn't break what we've done.

I still favour an approach that doesn't introduce an additional heading here. I don't think "Services List" adds any value to users, and we're only introducing this to get round a challenge in the CMS. If you were constructing a page like this in HTML, there is no reason you would introduce a h2 in this place.

I don't know enough to know if it's worse to have a skipped heading level than to have an unnecessary reading out of "Services List".

One thing worth noting is that "Services List" might not make sense in context if the pages linked from a Service Landing Page aren't in fact services.

markconroy:
> I don't think "Services List" adds any value to users, and we're only introducing this to get round a challenge in the CMS. 

I disagree. At the moment we have a list of teasers with no context to say what that list of teaser is about. A heading should be present to say "Most popular services" or something like that. This is not just getting around the limitations of the CMS, it's also providing screenreader users with an understandable hierarchy of the page. Sighted users may not need it as much as they can see the context from looking at the page.

The actual wording "services list" or whatever can easily be made configurable in the CMS.

msayoung:
Yep, as a follow up lets make that title customisable and with a flag whether it should be visually hidden or not. 

I agree that a heading is useful in this context.

msayoung (approved):

keelanfh:
Would the intention be to have it as a field on each Service Page, or globally for a site?

PR #420, opened 2023-03-23
sets banner image to aria-hidden true

### LINKED ISSUE #419: Should hero banner be marked as "decorative"
This came up during testing with a client:

> Can the homepage hero be decorative/skipped alt attribute?
>
> Due to the nature of the banner image against the page this should never hold value to understanding context of page, therefore should be skipped so it's not distracting.

What are our thougths?

markconroy:
Closes #419


msayoung:
comments #419

finnlewis:
Leaving this for now at Merge Monday, allowing discussion on #419 to come to a conclusion

markconroy:
I'm going to close this. Looks like we won't be using this PR, and will look to fix this if it's needed by adding a field on the paragraph type.

Issue #423, opened 2023-04-03
Incorrect display of list-item markers when links are inside <li>


nbrettell:
Setting all anchor links to inline-block display has caused the display of markers (bullet points) to appear at the bottom of multi-line list items. This might not have been the intended purpose.

![Demo site - with visible marker positions](https://user-images.githubusercontent.com/2014068/229474467-d676b27b-5cae-4b66-b1cf-fb008a13f0f7.png)

This was caused by the following commit: [changes to css \<a\> display](https://github.com/localgovdrupal/localgov_base/commit/ae8e922d91c1f8650d5813909e37429d86f9c025)

markconroy:
Related issue: https://github.com/localgovdrupal/localgov_base/issues/426

markconroy:
It might be worth reverting https://github.com/localgovdrupal/localgov_base/pull/408 (part of that was because of logo not displaying properly in some cases on Chrome), but there's another PR that (I think) fixed that, so maybe best to revert setting all links to `inline-block` by default.

markconroy:
@keelanfh wanna test https://github.com/localgovdrupal/localgov_base/pull/429

PR #424, opened 2023-04-03
Fix regression of list item bullet point position - issue 423

nbrettell:
To resolve display of list item markers (bullet points) for #423

### DIFF
--- a/css/base/base.css
+++ b/css/base/base.css
@@ -62,10 +62,6 @@ a:not(.toolbar a):focus {
   box-shadow: 0 -2px var(--color-focus), 0 4px var(--color-black);
 }
 
-ul li a {
-  display: block;
-}
-
 h1,
 .h1,
 h2,
--- a/css/base/base.css
+++ b/css/base/base.css
@@ -62,6 +62,10 @@ a:not(.toolbar a):focus {
   box-shadow: 0 -2px var(--color-focus), 0 4px var(--color-black);
 }
 
+ul li a {
+  display: block;
+}
+
 h1,
 .h1,
 h2,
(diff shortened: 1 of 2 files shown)


css/base/base.css
```diff
@@ -62,6 +62,10 @@ a:not(.toolbar a):focus {
   box-shadow: 0 -2px var(--color-focus), 0 4px var(--color-black);
 }
 
+ul li a {
+  display: block;
+}
```
markconroy: I think this `display: block` is a little too blunt here. It means all links in all lists use `block` which will be problematic if the links are within a string of text.

I think it would be better to set `block` on the specific items where we need it, such as the guides navigation.

nbrettell: Yeah I see what you mean, it would definitely cause a problem for strings of text. So this change wouldn't be satisfactory.

The original commit does disregard any other full links within list items throughout the site, which is obviously not as common

finnlewis: Just reviewing in Merge Monday,  any other ideas @nbrettell or @markconroy ?

markconroy: Hi @nbrettell 

Can you update your PR to remove your `display: block' and then in **guide-nav.css** change this:

```
.lgd-guide-nav__list-item > a {
  font-weight: bold;
}
```

to this
```
.lgd-guide-nav__list-item > a {
  display: inline;
  font-weight: bold;
}
```

That seems to fix the issue. If you get that change made, we can merge this then.

nbrettell: Hi @markconroy, thanks for the above information and apologies for not responding sooner. I have updated the PR

markconroy (approved):

PR #428, opened 2023-04-21
stop the messages block from being cached

### LINKED ISSUE #244: Persistent status message on node page
The node add/edit feedback message becomes persistent when that node is reloaded.  This happens with the localgov_base theme and its subthemes.

## Reproduction steps
- [Optional] Install a fresh localgov site.
- Login and create a node such as a Service page.
- As expected, the status message will say *Service page PAGE-TITLE has been updated.*
- Reload the newly created node while logged in as the same user.
- The status message appears again.  Same happens following further page reloads.
- Logout and reload the above page.  There are no status messages which is good.

markconroy:
Closes #244


ekes:
This should work out of the box? The block itself does 

```
  /**
   * {@inheritdoc}
   */
  public function getCacheMaxAge() {

    // The messages are session-specific and hence aren't cacheable, but the
    // block itself *is* cacheable because it uses a #lazy_builder callback and
    // hence the block has a globally cacheable render array.
    return Cache::PERMANENT;
  }
```
https://api.drupal.org/api/drupal/core%21modules%21system%21src%21Plugin%21Block%21SystemMessagesBlock.php/class/SystemMessagesBlock/8.1.x

What's the lazy_builder send?

stephen-cox:
This is a bit of a strange one as it should work out of the box, but doesn't, and the change here isn't altering the cache max age and so doesn't fix things for me.

Turning on cache debugging, on page load I'm seeing:

```
<!-- START RENDERER -->
<!-- CACHE-HIT: No -->
<!-- PRE-BUBBLING CACHE TAGS:
   * block_view
   * config:block.block.localgov_messages_scarfolk
-->
<!-- PRE-BUBBLING CACHE CONTEXTS:
   * languages:language_interface
   * theme
   * user.permissions
-->
<!-- PRE-BUBBLING CACHE KEYS:
   * entity_view
   * block
   * localgov_messages_scarfolk
-->
<!-- PRE-BUBBLING CACHE MAX-AGE: -1 -->
<!-- RENDERING TIME: 197.563431978 -->
<!-- START RENDERER -->
<!-- CACHE-HIT: Yes -->
<!-- CACHE TAGS:
   * block_view
   * config:block.block.localgov_messages_scarfolk
-->
<!-- CACHE CONTEXTS:
   * languages:language_interface
   * theme
   * user.permissions
-->
```

Comparing this to another site, it's loosing it's cache keys. Something's going wrong with the cache somewhere, but I haven't been able to track it down yet.

andybroomfield:
I suspect it is something to with [preprocessing to determine if there is something there](https://github.com/localgovdrupal/localgov_base/blob/9bdfe0c9f668d85de3748d8b85cdea973a4235b0/localgov_base.theme#L48)

Maybe we should just always print messages, Is it just the placeholder that is output and no extra markup?

stephen-cox:
@andybroomfield Your suspicion that the preprocess check was causing this issue is correct.

Rendering the block to check if there is any content is messing with the cache settings: https://github.com/localgovdrupal/localgov_base/blob/9bdfe0c9f668d85de3748d8b85cdea973a4235b0/localgov_base.theme#L61

I've created a new PR #439 that skips this for messages. This means that we now always end up rendering the divs that wrap messages as `has_messages` no longer exists and `{% if page.messages %}` is always true.

@markconroy I'm not sure how bad this is?

stephen-cox (commented):
This doesn't work for me.

I managed to fix this issue with https://github.com/localgovdrupal/localgov_base/pull/439, but this adds the div wrapping system messages to all pages and not sure if this is acceptable.

ekes:
Unless it's possible to remove the div's in the template. A template that doesn't contain anything else it's just the placeholder for content that gets loaded later. I think it is a choice between not knowing if it's there, and hence always having it; or knowing it's there but having already caused a render.

markconroy:
I'm closing this, since #439 has been merged.

PR #431, opened 2023-04-21
styles the skip to content link

### LINKED ISSUE #373: Skip link styling
## Proximity to edge of the window

The text in the skip link is right up against the edge of the window, which isn't ideal for legibility. Some spacing around this would be helpful. GOV.UK does this nicely (though I'm not suggesting that we use this overall styling - see below).

image

## Underline

There’s no visible thick underline on the component in the focus state, as there is for other `<a>` elements on the site.

In our designs, we're using a white header. If this was a normal component, the current focus indicator wouldn't be considered sufficient, as the yellow doesn’t have sufficient contrast against white, so we'd need the underline too. This element never appears in an unfocused state, so I'm not sure if this is absolutely required, but it might be good to keep it consistent with other `<a>` elements. This would mean that regardless of the colour of the header (councils will use different ones), the focus state is still clearly visible.

image

Localgovdrupal.org does this well - the focus indicator is clearly visible even with a yellow header:

image

tl;dr: I'm suggesting that the underline on the focus state of the skip link should be visible as 4px consistent with the other focus styles.

markconroy:
Closes #373

### DIFF
--- a/css/components/skip-link.css
+++ b/css/components/skip-link.css
@@ -1,0 +1,9 @@
+.skip-link {
+  padding: var(--spacing);
+}
+
+.skip-link:focus,
+.skip-link:hover {
+  text-decoration: underline !important;
+  text-decoration-thickness: max(3px, 0.1875rem, 0.12em) !important;
+}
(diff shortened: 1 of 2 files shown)


keelanfh (approved):
Looks great, big improvement and will work fine with our white header too.

msayoung (approved):

PR #432, opened 2023-04-21
removes references to Google fonts

### LINKED ISSUE #365: Subtheming creates links to Google Fonts
In https://github.com/localgovdrupal/localgov_base/blob/1.x/scripts/subtheme-items/templates/layout/_html.html.twig#L17-L19 there are some references to Google Fonts. This means that when you subtheme localgov_base, you end up with a page that makes requests to Google Fonts, even if you don't use this font - unless you remove these lines from the file.

Given [rulings on GDPR and Google Fonts](https://www.theregister.com/2022/01/31/website_fine_google_fonts_gdpr/), could we perhaps remove these, either by bundling a font in or defaulting to a system font?

markconroy:
Closes #365

### DIFF
--- a/scripts/subtheme-items/css/variables.css
+++ b/scripts/subtheme-items/css/variables.css
@@ -34,6 +34,6 @@
 */
 body {
   --color-accent: green;
-  --font-secondary: oswald, var(--font-primary);
+  --font-secondary: georgia, var(--font-primary);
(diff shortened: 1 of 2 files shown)


keelanfh (approved):
Fixes issue - and looks fine to me for a subtheme default. I expect most people will override this anyway.

PR #436, opened 2023-05-02
Add extend for cookie compliance module

MatthewWilkes:
This extends the `localgov_eu_cookie_compliance` module to add the additional CSS to integrate it with the base theme.

This is a counterpart to https://github.com/localgovdrupal/localgov_eu_cookie_compliance/pull/4

### DIFF
--- a/css/components/localgov_eu_cookie_compliance.css
+++ b/css/components/localgov_eu_cookie_compliance.css
@@ -1,0 +1,18 @@
+/*
+  @file Theming for localgovdrupal/localgov_eu_cookie_compliance
+  
+  This has been included to avoid the cookie compliance integration
+  having a dependency on the base theme's variables for styling. By
+  including this here, there will be sensible behaviour out-of-the-box.
+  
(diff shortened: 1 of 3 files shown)


MatthewWilkes:
Have marked as ready after confirming that this doesn't cause issues if `localgov_base` is installed but `localgov_eu_cookie_compliance` is not. The library being merged in _must_ exist, but it's only merged if the target is present.

css/components/localgov_eu_cookie_compliance.css
```diff
@@ -0,0 +1,18 @@
+/*
+  @file Theming for localgovdrupal/localgov_eu_cookie_compliance
+  
+  This has been included to avoid the cookie compliance integration
+  having a dependency on the base theme's variables for styling. By
+  including this here, there will be sensible behaviour out-of-the-box.
+  
+  End-users can override these styles in their decendent themes, if needed.
```
Polynya: Typo - descendant

PR #439, opened 2023-05-03
Prevent system messages from being cached

### LINKED ISSUE #244: Persistent status message on node page

stephen-cox:
Don't check if there are messages in preprocess hook 

Fixes #244

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -52,14 +52,12 @@ function localgov_base_preprocess_page(&$variables) {
     $excluded_regions = [
       'messages',
     ];
+    if (in_array($region, $excluded_regions)) {
+      continue;
+    }
 
-    if (!in_array($region, $excluded_regions)) {
-      $copy = $variables['page'][$region];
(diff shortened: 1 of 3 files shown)


stephen-cox:
Further discussion of the implications of this change on this comment: https://github.com/localgovdrupal/localgov_base/pull/428#issuecomment-1532965127

andybroomfield:
This is the same fix BHCC has added.

finnlewis:
@ekes would rather wait for @markconroy to review this to confirm markup is OK.

markconroy (approved):
Looks good to me

PR #443, opened 2023-05-19
adds heading and skip link for guide navigation

### LINKED ISSUE #433: Guide navigation repeated - repeating unnecessary content for screenreader users
We recently did a round of testing with screenreader users. 

Generally, users found the guide navigation helpful.

When clicking on links in the guide navigation, two out of the four screenreader users we tested with expected the link to bring them directly to a heading and the content of that guide section - rather than repeating the guide navigation section again. 

One user commented that it would be helpful to have a heading for the guide navigation - if they're navigating using headings, there isn't any way of seeing that the guide navigation is there. If you look at [guides on gov.uk](https://www.gov.uk/bereavement-support-payment/how-to-claim), there's a heading ("Contents") above the guide navigation. It could be worth including at least a visually-hidden heading here to assist screenreader users with navigation.

It could also be worth adding a link to skip the guide navigation - this wasn't asked for by any of our users, but could be helpful, again for avoiding repeating this. At the moment I think we meet [2.4.1 Bypass Blocks](https://www.w3.org/WAI/WCAG21/Understanding/bypass-blocks.html) through the use of the navigation landmark, but we did find that not all screenreader users were familiar with navigation landmarks and wouldn't necessarily use this to skip the repeated content.

Somewhat related to #416.

markconroy:
Closes #433

### DIFF
--- a/templates/block/guides-contents-block.html.twig
+++ b/templates/block/guides-contents-block.html.twig
@@ -16,6 +16,8 @@
     {% else %}
       {% set list_type = 'ul' %}
   {% endif %}
+  <h2 class="visually-hidden">{{ 'Guide Navigation'|t }}</h2>
+  <a class="visually-hidden" href="#lgd-guides__title">{{ 'Skip Guide Navigation'|t }}</a>
   <nav class="lgd-guide-nav" aria-label="{{ 'Guide'|t }}">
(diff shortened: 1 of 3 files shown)


keelanfh:
Tested in VoiceOver on Mac and works as expected.

keelanfh (approved):

PR #444, opened 2023-05-19
links prev/next in guides to content instead of page

### LINKED ISSUE #416: Link guide prev/next to the guide content
At the moment when you click on prev/next for a guide, it directs to that guide page, with focus at the top of the page.

Should we link to the `lgd-guides__inner` section instead so you are brought to the next "section" of the guide rather than the next "page" of the guide?

markconroy:
Closes #416

### DIFF
--- a/templates/block/guides-prev-next-block.html.twig
+++ b/templates/block/guides-prev-next-block.html.twig
@@ -5,7 +5,7 @@
 */
 #}
 
-{% 
+{%
   set classes = [
     'lgd-prev-next',
     'lgd-prev-next--guides',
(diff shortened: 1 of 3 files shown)


keelanfh:
Tested without any assistive tech and this worked fine.

However, testing with VoiceOver switched on (Chrome/Safari, macOS), when clicking on the link, VoiceOver started reading from the "Skip to main content" at the top of the page was read out, which negates the benefit for these users. Any idea why that's happening or if there's something we can do about that?

finnlewis:
Leaving this for review.

finnlewis (commented):
Tested locally and works for me in the browser. 

@keelanfh I've not tested with a voice over, are you using an extension for that or something native?

finnlewis (approved):

keelanfh:
I used VoiceOver on Mac (it's native). We can just create a separate issue to have a look at that, this is still an improvement just not for people using that.

keelanfh (approved):

PR #445, opened 2023-05-19
adds aria label to services menu button

### LINKED ISSUE #411: Services menu: moving focus in to the sub-menu
Using a screen reader (NVDA) to navigate the site - 
Tab to the Services menu button , the screen reader reads "Services button collapsed"
I would expect to press the button and hear "Services button expanded" but instead when I press the button focus moves to the first item  in the list. 

My concern is that a person with sight-loss will not know that the pressing of the button expanded the menu, not hearing the expected "button expanded". 

https://www.awesomescreenshot.com/video/15694110?key=df4ffa6a247bc3b5b401b456efc605af

## Recommendation
Leave focus on the button, which will read the aria-expanded=true attribute. Let users move in to the expanded menu of their own accord. 

## Update
I recall why this is complicated. The services menu itself in not semantically immediately after the services button, instead its in a region after the header. In this case maybe we need to consider better (aria) labelling instead. 
Something like: `aria-label="Services: expand and jump to services menu"`

@danchamp and others, I wonder if you have thoughts on 1. the severity of the issue and 2. a good solution ? 
Are there any related accessibility issues we should consider alongside this?

markconroy:
Closes #411

### DIFF
--- a/templates/layout/header.html.twig
+++ b/templates/layout/header.html.twig
@@ -25,6 +25,7 @@
                   data-target="lgd-header__nav--secondary"
                   aria-controls="lgd-header__nav--secondary"
                   aria-expanded="false"
+                  aria-label="{{ 'Services: expand and jump to services menu'|t }}"
                 >
                   <span class="lgd-header__toggle-text lgd-header__toggle-text--secondary">{{ 'Services'|t }}</span>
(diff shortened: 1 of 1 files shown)


danchamp (approved):
Looks good to me.

PR #449, opened 2023-05-26
sets guide nav skip link to be focusable

### LINKED ISSUE #448: Make guide navigation skip link visible when focused
Following on from #443 

Can we make the "Skip guide navigation" link visible when focused, like we'd do for the "Skip to main content" link?

At the moment, it captures focus but you can't see anything so you can "lose" where the focus is as you tab through the page

markconroy:
Closes #448


keelanfh (approved):
LGTM 🎉

PR #451, opened 2023-06-02
focuses guide title when linking from guide prev/next

### LINKED ISSUE #450: Guides: Previous/Next button to go to main content for VoiceOver users
Following on from #444

Previous/Next button when VoiceOver on Mac is switched on still brings users to the top of the page rather than the main content is intended.

Since one of the reasons for this change was to benefit screenreader users, we should take a look at this and see if we can change something that it works for VoiceOver users too. We could also take a look at this with NVDA or JAWS and see if we observe similar behaviour.

markconroy:
Closes #450

### DIFF
--- a/localgov_base.libraries.yml
+++ b/localgov_base.libraries.yml
@@ -124,7 +124,7 @@ topic-list-builder:
     theme:
       css/components/topic-list-builder.css: {}
 
-guide:
+guides:
   js:
     /js/guides.js: {}
 
--- a/localgov_base.libraries.yml
+++ b/localgov_base.libraries.yml
@@ -124,6 +124,10 @@ topic-list-builder:
     theme:
       css/components/topic-list-builder.css: {}
 
+guide:
+  js:
+    /js/guides.js: {}
+
 guide-nav:
   css:
     theme:
--- a/js/guides.js
+++ b/js/guides.js
@@ -1,0 +1,24 @@
+/**
+ * @file JS file for the Guides components.
+ */
+
+(function guidesScript(Drupal) {
+  Drupal.behaviors.guides = {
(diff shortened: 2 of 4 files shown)


keelanfh:
This doesn't seem to work for me, not sure if I'm doing something wrong, but I'm seeing it work just the same as before. Does it work for you with VoiceOver?

localgov_base.libraries.yml
```diff
@@ -124,6 +124,10 @@ topic-list-builder:
     theme:
       css/components/topic-list-builder.css: {}
 
+guide:
```
keelanfh: should this be 'guides'? or else change in the attach_library so that it matches?

finnlewis:
Leaving this for @markconroy to review

keelanfh:
Still doesn't work for me - I'm not sure if I'm just not clearing caches properly or something? Can someone else test it?

markconroy:
@keelanfh what's not working exactly? As in, when you click "next" and go to the next guide page, what is set as the focussed item? 

In my testing, it was the title. So if I hit tab I then go to the first link inside that guide item, and if it hit Shift+Tab I go to the last item in the guide nav. Is this not happening for you?

keelanfh:
No... for me when I hit the next button, it's just bringing focus to the "Skip to main content" link... that's why I'm wondering whether I've applied the patch correctly, as I can't see the javascript in there. I've taken the same approach as before, but happy if you want to merge/release and then I can look at it again

keelanfh (approved):

finnlewis:
@markconroy shall we go ahead and merge and release this?

markconroy:
Yes, I think so.

PR #453, opened 2023-06-16
fixes header toggles not working when empty regions

### LINKED ISSUE #452: If primary nav/search is empty OR secondary nav is empty, header toggles fail
The JS for the header toggles creates a variable called `const region = ...`, if there's nothing in the regions that the primary/secondary toggle looks for, then we can "region is null" error and the rest of the JS in that file doesn't load.

a simple `if (region) { do stuff }` should suffice to fix it, around line 48 (before we declare `firstlink`)
https://github.com/localgovdrupal/localgov_base/blob/1.x/js/header.js#L48

markconroy:
Closes #452 

To test:
- Remove blocks from Search and Primary Nav regions, so that region will not be rendering
- The "Services" toggle should stop working as the JS will break looking for a 'primary' key for 'region'
- Add this PR which does a check for primary before calling it's code, skips it when it's not found, and continues to work

### DIFF
--- a/js/header.js
+++ b/js/header.js
@@ -45,9 +45,10 @@
           : 'secondary';
         // The resulting region.primary.firstLink isn't used, but it's less
         // difficult to add it than to add only region.secondary.firstLink.
-        const firstLink = region.querySelector('.menu a');
-
-        navInfo[nav] = { toggle, region, firstLink };
+        if (region) {
+          const firstLink = region.querySelector('.menu a');
+          navInfo[nav] = { toggle, region, firstLink };
+        }
       });
 
       // When a menu toggle button is clicked, show/hide the menu regions.
@@ -138,7 +139,7 @@
             navInfo.primary.toggle.addEventListener('click', handlePrimaryMenuToggleClick);
           }
         } else {
-          if (navInfo.primary.toggle) {
+          if (Object.keys(navInfo).includes('primary') && navInfo.primary.toggle) {
             navInfo.primary.toggle.removeEventListener('click', handlePrimaryMenuToggleClick, true);
           }
           if (Object.keys(navInfo).includes('secondary') && navInfo.secondary.toggle) {
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -5,9 +5,9 @@
  * Theme hooks to support the LocalGov Base Theme.
  */
 
+use Drupal\Component\Utility\Crypt;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\views\ViewExecutable;
-use Drupal\Component\Utility\Crypt;
 
 /**
  * Implements hook_form_system_theme_settings_alter().


js/header.js
```diff
@@ -138,7 +139,7 @@
             navInfo.primary.toggle.addEventListener('click', handlePrimaryMenuToggleClick);
           }
         } else {
-          if (navInfo.primary.toggle) {
+          if (Object.keys(navInfo).includes('primary') && navInfo.primary.toggle) {
```
ctorgalson, line 142:
```suggestion
          if (navInfo.primary?.toggle) {
```

It's also good as-is. This just has the same effect while being a bit more readable.

js/header.js
```diff
@@ -45,9 +45,10 @@
           : 'secondary';
         // The resulting region.primary.firstLink isn't used, but it's less
         // difficult to add it than to add only region.secondary.firstLink.
-        const firstLink = region.querySelector('.menu a');
-
-        navInfo[nav] = { toggle, region, firstLink };
+        if (region) {
+          const firstLink = region.querySelector('.menu a');
+          navInfo[nav] = { toggle, region, firstLink };
+        }
```
ctorgalson, line 51: I'm confused about how the situation this addresses could come about.

For the `if (region) { ...` check to be useful, it has to mean that there *is* a toggle with an `aria-controls` attribute, but there *is not* a corresponding element with a matching `id` attribute.

So while this change will *work*, it could cause a confusing situation in those screenreaders that support `aria-controls`.

This makes me think a better solution would be to make sure that `toggle` can't exist if `region` doesn't (possibly in Twig)? But I see the template is checking [`{% if has_primary_menu or has_search %}`](https://github.com/localgovdrupal/localgov_base/blob/639c927c692850c52e00a6b8765895378e1b33fc/templates/layout/header.html.twig#L35) so I'm not sure if this is an option.

Alternatively something like this could address the `aria-controls` issue without resolving the _actual_ problem (and note we can't just remove `aria-controls` in case we add the search id to it):

```suggestion
        if (region) {
          const firstLink = region.querySelector('.menu a');
          navInfo[nav] = { toggle, region, firstLink };
        } else {
          toggle.setAttribute(
            'aria-controls',
            toggle.getAttribute('aria-controls').replace(toggle.id, '').trim()
          );
        }
```

ctorgalson (commented):
Second change is definitely fine, but the first looks like something that should only be handled in javascript if no other fix is possible--I can't tell for sure (but I also see that the primary toggle's `aria-controls` attribute contains no reference to the `search` region.

js/header.js
ctorgalson: Just ran across this tab again: I didn't spot it at first, but even without resorting to `primary?`, it's still possible to simplify `Object.keys()` plus `includes()`. The js object prototype has a method specifically for this, `hasOwnProperty()`:

```suggestion
          if (navInfo.hasOwnProperty('primary') && navInfo.primary.toggle) {
```

ctorgalson (approved):
Fixes current issue. Save further jiggery-pokery for when we have another go at header.js.

PR #461, opened 2023-06-29
Add alert-success style

### LINKED ISSUE #460: WYSIWYG alert styles are incorrect
`.alert-success` is missing from wysiwyg-styles in localgov_base, falling back to `.alert` which could be the wrong colour/impression

in the wysiwyg config `p.alert.alert-primary|Alert failure` but there is .alert-fail in the CSS. 

.alert-fail {
  border-color: var(--color-warning); 
}

image

keelanfh:
Looks like alert-success style was missing, so let's add them.

I don't know why `ckeditor.css` doesn't use variables, but I assume there's a good reason, so I've followed the same pattern.

Fixes #460

### DIFF
--- a/css/base/ckeditor.css
+++ b/css/base/ckeditor.css
@@ -30,6 +30,10 @@
   border-color: #ed7522;
 }
 
+.alert-success {
+  border-color: #048a04;
+}
+
 /* Links */
 .btn.btn-start {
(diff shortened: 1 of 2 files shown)


stephen-cox (approved):
This looks good to me.

PR #462, opened 2023-06-30
Correct styles and height of CKEditor 5 WYSIWYG

keelanfh:
To support upgrade to Drupal 10, this covers:
* adding a new CSS file for CKEditor 5 so styles appear in the WYSIWYG
* declaring this file in the info.yml
* some tweaks to take account of differences in the new CKEditor 5
* adding a minimum height to the CKEditor boxes

With thanks to @justinepocock for advice in https://github.com/localgovdrupal/localgov_core/issues/181#issuecomment-1600902343.

### DIFF
--- a/css/base/ckeditor5.css
+++ b/css/base/ckeditor5.css
@@ -1,0 +1,105 @@
+/*
+  @file Styles for ckeditor5.
+*/
+
+/* Alerts */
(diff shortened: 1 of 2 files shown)


justinepocock (approved):
Looks good to me

PR #463, opened 2023-07-04
Enhance focus states for input, textarea and select #459

### LINKED ISSUE #459: Enhance the focus states on form inputs
Form inputs just have default focus states, not the enhanced GOV.UK design system ones like we see in other places. It would be good to apply the GOV.UK design system ones here too.

keelanfh:
Adding GDS-style focus states for input, textarea and select fields.

I've left the border in the unfocused state the same but just applied the enhanced focus indicators.

This should make it more consistent with the other focus states for radios and checkboxes.

Closes #459.

### DIFF
--- a/css/components/form-items.css
+++ b/css/components/form-items.css
@@ -63,7 +63,7 @@ input[type="text"]:focus,
 input[type="url"]:focus,
 textarea:focus,
 select:focus {
-  outline: 3px solid var(--color-yellow);
+  outline: 3px solid var(--color-focus);
   outline-offset: 0;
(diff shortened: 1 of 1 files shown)


css/components/form-items.css
```diff
@@ -63,7 +63,7 @@ input[type="text"]:focus,
 input[type="url"]:focus,
 textarea:focus,
 select:focus {
-  outline: 3px solid #fd0;
+  outline: 3px solid var(--color-yellow);
```
keelanfh: Maybe `--color-focus` instead?

Issue #349, opened 2023-01-20
Let's stop explicitely supporting IE11 for the frontend theme


markconroy:
- We have ie11 CSS files in our codebase. Keeping them up-to-date when editing the theme is cumbersome. 
- If you look at an LGD site in IE11, it's pretty ugly (because it uses the default fallback colours rather than the new colours from your custom theme).
- Since no one has complained about "our site uses ugly colours in IE11" I'm left thinking of two possibilities
- - People really like our IE11 design
- - No one has visited our LGD sites in IE11 (or maybe they are just used to sites looking ugly)
- If the latter is true, we are maintaining something that is not of value to our users
- Debugging takes longer with IE11 being supported, because everytime you want to comment out some code, you may also need to comment out the IE11 code). This is especially annoying when working in the inspector.
- If we drop IE11 support, we can rewrite our grid system in grid instead of flexbox and this will have the benefit of removing many of the annoyances of 8px left/right margins we have that mean things almost line up perfectly with each other, but not fully. 
- If we keep supporting IE11, we have to keep using part of JS that are no modern or use polyfills instead or compilers, etc, rather than writing ES6+ JS.
- If we drop IE11 support, we will have about 30% less code in localgov_base
- We'll be dropping IE11 support soon in any case when we move to D10, so I'm proposing we do so now instead.

keelanfh:
+1 for this from Essex, in general.

Our analytics over the last 30 days shows <0.002% of our users using IE - 5 visitors out of 293,000.

If the stats on caniuse.com are accurate, I'd guess we exclude about 500x as many users from accessing our site entirely through not supporting anything less than TLS v1.2 - as does gov.uk.

Question @markconroy - would dropping support mean that this content would be invisible/obscured in IE11? Or just styling wouldn't be as we expect? It looks like GOV.UK dropped support to a 'functional' level: https://github.com/alphagov/govuk-frontend/discussions/2607. I'm not too bothered either way, just curious.

markconroy:
It just means the styling wouldn't be as expected. You'd get some styling if IE11 can use those properties, so we wouldn't be actively trying to break things, just not worrying about them in the first place.

finnlewis:
This is interesting: 

https://technology.blog.gov.uk/2022/06/16/service-manual-testing-requirement-changes-for-internet-explorer-11/

Let's ask our councils for data to hopefully support this decision. 

[...]

PR #467, opened 2023-07-07
Makes default link focus style less specific

### LINKED ISSUE #466: Base link :focus style too specific, overrides other styles
## Summary

We're finding [the `localgov_base` link focus style `a:not(.toolbar a):focus`](https://github.com/localgovdrupal/localgov_base/blob/1.x/css/base/base.css#L54) unnecessarily difficult to override due to its scope and specificity.

## Explanation

In terms of _scope_, it applies to literally every link in any `localgov_base` child theme that isn't inside `.toolbar` regardless of whether the toolbar is even present. This is confusing and adds a little friction with every link style whose focus doesn't use `localgov_base`'s styles.

In terms of _specificity_, `:not()` has the specificity of its _argument_. This means that the base focus style will override something like `.call-out-box__link:focus` unless we scope it under a parent like `.call-out-box .call-out-box__link:focus` or override the css variable responsible for whatever the problematic property is.

For example, to override the focus background colour on a link, both of these strategies _do_ work:

```css
/* Scoped/descendant selector */
.call-out-box .call-out-box__link:focus {
  background: var(--color-white);
}
```

```css
/* Scoped variable override */
.call-out-box__link:focus {
  --color-focus: var(--color-white);
}
```

## Alternate strategy

Since the `:not()` in the selector is intended to address _only_ links inside `.toolbar`, it should be possible to:

1. change the selector to `.dialog-off-canvas-main-canvas a:focus` (thus targeting virtually everything _but_ `.toolbar`, **or**
2. remove the `:not()` from the selector and add a new rule with the selector `.toolbar a:focus` that specifically un-sets everything that the `a:focus` selector does.

Of the two, I'd suggest (1) since the only non-toolbar link it will miss is `.skip-link` which could be added specifically to the base `:focus` style.

On the other hand, using (2) means `localgov_base` has to take responsibility for the appearance of `.toolbar`, even when Toolbar's own css library changes.

## References:

- https://developer.mozilla.org/en-US/docs/Web/CSS/:not#description
- https://css-tricks.com/almanac/selectors/n/not/#aa-specificity

ctorgalson:
This change makes the selector less specific, though [it is still technically a bit more specific than e.g. `.call-out-box__link:focus`](https://polypane.app/css-specificity-calculator/#selector=a%3Anot(.toolbar%20a)%3Afocus%2C%20a%3Anot(%5Bclass*%3D%22toolbar%22%5D)%3Afocus%2C%20.call-out-box__link%3Afocus).

On the other hand, it seems close enough in specificity that [a selector with a simple class coming later in the cascade can override it as expected](https://codepen.io/bedlamhotel/pen/VwVMLBz).

This shouldn't interfere with whatever level of IE support the theme aims to maintain as [the `[class*="toolbar"]` selector has the same level of support as the original's `:not()`](https://caniuse.com/css-sel3).

Re: #466

### DIFF
--- a/css/base/base.css
+++ b/css/base/base.css
@@ -51,7 +51,7 @@ a:hover {
   text-decoration-thickness: max(3px, 0.1875rem, 0.12em);
 }
 
-a:not(.toolbar a):focus {
+a:not([class*="toolbar"]):focus {
   text-decoration: none;
   color: var(--color-black);
   outline: 3px solid transparent;
--- a/css/components/footer.css
+++ b/css/components/footer.css
@@ -12,9 +12,6 @@
 .lgd-footer__pre-footer a {
   color: var(--color-pre-footer-link);
 }
-.lgd-footer__pre-footer a:focus {
-  color: var(--color-grey-dark);
-}
 
 .lgd-footer__footer {
   padding-top: clamp(1rem, 10vw, var(--section-spacing-vertical-footer));


keelanfh:
The specificity change causes the footer link text to be grey in the focus state, rather than the previous black:

image

Not sure where this came from but it might have been from a previous design. I suggest removing this:

https://github.com/localgovdrupal/localgov_base/blob/cbecf8af27739a32882e5bf2af5c1071c513a3eb/css/components/footer.css#L15-L17

keelanfh (changes requested):
I've tested this and it looks good to me, apart from the one change I mentioned.

For future, you should be able to create branches in this repo, which makes it easier for us to collaborate on the same branch :)

keelanfh:
> This shouldn't interfere with whatever level of IE support the theme aims to maintain as [the [class*="toolbar"] selector has the same level of support as the original's :not()](https://caniuse.com/css-sel3).

Oh, and [we don't support IE any more](https://github.com/localgovdrupal/localgov_base/issues/349).

ctorgalson:
> I've tested this and it looks good to me, apart from the one change I mentioned.

I've pushed that up.

> Oh, and we don't support IE any more.

I thought not :partying_face:

keelanfh (approved):

markconroy:
Thanks @ctorgalson and @keelanfh

PR #481, opened 2023-09-04
422 show titles of previousnext pages in guide navigation buttonslinks

### LINKED ISSUE #422: Show titles of previous/next pages in guide navigation buttons/links
At the moment, the previous/next button on guide pages doesn't show any information about where the guide pages go:

image

Particularly for longer pages, this might mean that a user doesn't know what to expect when they click the link, especially on mobile devices - see example of a longer page below - and we might be expecting them . @helencalderon reminded me that this is an example where we can follow the [Recognition over Recall usability heuristic](https://www.nngroup.com/articles/ten-usability-heuristics/#:~:text=%236%3A%20Recognition%20rather%20than%20recall).

@markconroy did [some work](https://github.com/essexcountycouncil/send-lgdrupal/blob/main/web/themes/custom/anrt_lgd/templates/block/guides-prev-next-block.html.twig) on this for send.essex.gov.uk, which brings this more in line with the existing essex.gov.uk and the way that GOV.UK does it. Mark, if this was something that others wanted in LGD, I don't know how long this would take to move across?

image

Example of this interaction in context:

alexburrows:
Notification_Center

This is based off @markconroy css / markup approach and has been modified to allow the settings option we have via the block

### DIFF
--- a/css/components/guide-nav.css
+++ b/css/components/guide-nav.css
@@ -37,25 +37,3 @@
   margin-top: var(--spacing-largest);
   margin-bottom: var(--spacing-largest);
 }
-
-.lgd-prev-next__link {
-  flex-wrap: wrap;
-  padding: var(--spacing-smaller);
-  text-decoration: none;
-}
-
-.lgd-prev-next__label {
-  font-weight: bold;
-}
-
-.lgd-prev-next__icon {
-  margin-top: var(--spacing-smallest);
-}
-
-.lgd-prev-next__title {
-  width: 100%;
-}
-
-.lgd-prev-next__list-item--prev .lgd-prev-next__title {
-  padding-left: var(--spacing-large);
-}
--- a/css/components/guide-nav.css
+++ b/css/components/guide-nav.css
@@ -24,3 +24,42 @@
   display: inline;
   font-weight: bold;
 }
+
+.lgd-guide-nav__list-item--active {
+  font-weight: bold;
+}
+
+.lgd-guide-nav__list-item > a {
+  font-weight: normal;
+}
+
+.block-localgov-guides-contents {
+  margin-top: var(--spacing-largest);
+  margin-bottom: var(--spacing-largest);
+}
+
+.lgd-prev-next__link {
+  flex-wrap: wrap;
+  padding: var(--spacing-smaller);
+  text-decoration: none;
+}
+
+.lgd-prev-next__link  > .lgd-prev-next__label {
+  font-weight: bold;
+}
+
+a.lgd-prev-next__link .lgd-prev-next__label {
+  font-weight: bold;
+}
+
+.lgd-prev-next__icon {
+  margin-top: 3px;
+}
+
+.lgd-prev-next__title {
+  width: 100%;
+}
+
+.lgd-prev-next__list-item--prev .lgd-prev-next__title {
+  padding-left: var(--spacing-large);
+}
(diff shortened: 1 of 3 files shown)


css/components/guide-nav.css
```diff
@@ -24,3 +24,42 @@
...
+.block-localgov-guides-contents {
+  margin-top: var(--spacing-largest);
+  margin-bottom: var(--spacing-largest);
+}
+
+.lgd-prev-next__link {
+  flex-wrap: wrap;
+  padding: var(--spacing-smaller);
+  text-decoration: none;
+}
```
markconroy: Can you check that this `flex-wrap` works? Without a `display: flex`, I'm not sure it will.

css/components/guide-nav.css
```diff
@@ -24,3 +24,42 @@
...
+
+.lgd-prev-next__link  > .lgd-prev-next__label {
+  font-weight: bold;
+}
+
+a.lgd-prev-next__link .lgd-prev-next__label {
+  font-weight: bold;
+}
```
markconroy: Can this all be reduced to something like:
```
.lgd-prev-next__label {
  font-weight: bold;
}
```

This looks like it's it's the same element being set to `bold` twice.

css/components/guide-nav.css
```diff
@@ -24,3 +24,42 @@
...
+
+.lgd-prev-next__icon {
+  margin-top: 3px;
```
markconroy: Let's use a variable here such as `--spacing-smallest`

markconroy (commented):
Thanks for this Alex, looks really good. I've just a few teeny, weeny requests and then we're ready to merge.

css/components/guide-nav.css
finnlewis: @alexburrows , just discussing at Merge Monday and @markconroy just wanted to check on the flex-wrap question.

finnlewis:
How's this looking @markconroy ?

markconroy:
Same as last week, we're still waiting on @alexburrows to verify the `flex-wrap` item if there is no `display: flex` for that item.

If we don't want to wait for that, I'm happy for this to be merged and we can file a follow-up/clean-up issue if we need it later.

alexburrows:
sorry comment showed pending, so the code without it in place breaks the alignment. Where as how we have it now works

finnlewis:
Sorry @alexburrows , could you confirm that `flex-wrap` is intended on `.lgd-prev-next__link` ?

https://github.com/localgovdrupal/localgov_base/pull/481/files#diff-739d0a730f224dcf99050f50bd6b64923dfd1b866f46a784e67617353d167ae1R41-R42

I think @markconroy 's comment was asking if this had any effect without `display: flex`

> 
> Member
> @[markconroy](https://github.com/markconroy) markconroy [3 weeks ago](https://github.com/localgovdrupal/localgov_base/pull/481#discussion_r1315042472)
> Can you check that this flex-wrap works? Without a display: flex, I'm not sure it will.
>

markconroy:
I found the reason we had no `display: flex` - the `display: flex` is in the prev-next.css file and this PR puts `prev-next` classes into  the `guide-nav.css` file. 

I've moved the code to the correct file now. Any one want to give it another quick test? I think it's ready to merge.

stephen-cox (approved):
Looking good.

Only thing I suggest is turning the title on by default, but I'll open an issue in Guides for that.

PR #492, opened 2023-10-16
Makes page header block lede rendering slightly more flexible

### LINKED ISSUE #491: localgov_base directly access #value property of lede in localgov-page-header-block.html.twig
The [`localgov-page-header-block.html.twig` template accesses `lede['#value']` instead of printing `{{ lede }}`](https://github.com/localgovdrupal/localgov_base/blob/1.x/templates/block/localgov-page-header-block.html.twig#L30).

This is because of how `localgov_guides` creates an unmodifiable render array in [`src/EventSubscriber/PageHeaderSubscriber.php`](https://github.com/localgovdrupal/localgov_guides/blob/2.x/src/EventSubscriber/PageHeaderSubscriber.php#L45-L49).

localgovdrupal/localgov_guides#122 and the proposed PR localgovdrupal/localgov_guides#133 are planning to change their rendering of the block, so the current template is too inflexible. We can change this while retaining the current as a default. PR follows.

ctorgalson:
- Retains current rendering, but anticipates that the {{ lede }} render array could have a different structure.
- A forthcoming change to localgovdrupal/localgov_guides may require this.
- See: https://github.com/localgovdrupal/localgov_guides/pull/133
- Re: 491-localgov_base-directly-access-value-property-of-lede-in-localgov-page-header-blockhtmltwig

### DIFF
--- a/templates/block/localgov-page-header-block.html.twig
+++ b/templates/block/localgov-page-header-block.html.twig
@@ -26,8 +26,12 @@
       <h1 class="lgd-page-title-block__title">{{ title }}</h1>
     {% endif %}
 
-    {% if lede['#value'] %}
-      <p class="lgd-page-title-block__subheader">{{ lede['#value'] }}</p>
+    {% if lede %}
+      {% if lede['#value'] %}
+        <p class="lgd-page-title-block__subheader">{{ lede['#value'] }}</p>
+      {% else %}
+        {{ lede }}
+      {% endif %}
     {% endif %}
 
   </div>


finnlewis:
Thanks for this @ctorgalson 

Discussing in Merge Monday and think we need some input from Mark and perhaps look for consistency in the approach between guides, step-by-step and publications. 

In localgov_publications we've gone the NHS route: https://github.com/localgovdrupal/localgov_publications/blob/1.x/templates/localgov-publication-page-header-block.html.twig#L20-L31

Be interested to hear from @markconroy on this too.

andybroomfield:
Ref, also maybe a third option https://github.com/localgovdrupal/localgov_core/issues/192

markconroy:
I'm not fully sure of the problem being solved, but from a code perspective this looks "good enough for now; safe enough to try". We can create a follow up if we want/need to so that we have a unified approach to these type things.

ctorgalson:
@markconroy The idea is to make `localgov_base` less reliant on a render array with a specific structure.

Changes to `localgov_guides` _may_ change the array, but it's already possible for any sub-theme or module to preprocess `lede` into something that doesn't _have_ a `#value` property.

finnlewis:
Just discussing in Merge Monday. 

Andy mentioned https://github.com/localgovdrupal/localgov_core/pull/193 which I see you have comment on @ctorgalson 

I also see https://github.com/localgovdrupal/localgov_guides/pull/133#issuecomment-1794343128 

So.... actually this makes sens for now. @markconroy is happy to merge this for now, making the $lede variable more flexible in localgov _base.

finnlewis (approved):

PR #493, opened 2023-10-19
better contrast for links in quote paragraphs

### LINKED ISSUE #31: Insufficient contrast between --color-accent and grey background
The blue text in grey boxes does not contain enough contrast (3.49) between background and foreground colour.

I have added a screenshot for an example of said box on the Subsite example page for different page components.

https://test.localgovdrupal.org/test-subsite-demo-all-components/example-page-different-page-com

@msayoung 

hcm_scarfolk1

markconroy:
Closes https://github.com/localgovdrupal/localgov_scarfolk/issues/31


bisd (approved):
tested with change and new contrast is 4.53:1.

PR #498, opened 2023-11-01
Stops blocks that are assigned to the disabled region being rendered.

### LINKED ISSUE #497: Disabled blocks get rendered unnecessarily
Any blocks that are placed in the 'Disabled' region get rendered, even though they're never displayed.

rupertj:
Added 'disabled' region to the list of regions not to render. Limited region list to visible ones. Moved list of excluded region names out of the loop and simplified vars in the loop.

Fixes #497.

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -43,22 +43,20 @@ function localgov_base_preprocess_page(&$variables) {
   // @see https://www.drupal.org/node/953034
   // @see https://www.drupal.org/forum/support/module-development-and-code-questions/2016-04-07/drupal-8-regions-with-and-empty#comment-12149518
   $active_theme = \Drupal::service('theme.manager')->getActiveTheme()->getName();
-  $system_region = system_region_list($active_theme, $show = REGIONS_ALL);
+  $regions = array_keys(system_region_list($active_theme, REGIONS_VISIBLE));
 
-  foreach ($system_region as $key => $value) {
-    $region = $key;
-    $has = 'has_' . $region;
+  $excluded_regions = [
+    'messages',
+    'disabled',
+  ];
 
-    $excluded_regions = [
-      'messages',
-    ];
+  foreach ($regions as $region) {
     if (in_array($region, $excluded_regions)) {
       continue;
     }
-
     $copy = $variables['page'][$region];
-    $rendered = Drupal::service('renderer')->renderPlain($copy);
-    $variables[$has] = !empty(trim(strip_tags($rendered, '<drupal-render-placeholder><embed><hr><iframe><img><input><link><object><script><source><style><video>')));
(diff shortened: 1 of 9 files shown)


finnlewis:
Maybe want testing to confirm and approve

rupertj:
So I've added a test for this change. Turns out it was pretty hard to test...

There's a couple of fails showing up, but they're caused by this: https://github.com/localgovdrupal/localgov_project/issues/144, not the code in this PR.

rupertj:
Tests are fully green again. What do you reckon @finnlewis?

finnlewis:
@ekes is taking a look

tests/Functional/RegionRenderTest.php
```diff
@@ -0,0 +1,70 @@
...
+   * module into one of the regions. We then load a page, which is a callback
+   * provided by a controller in the same module, that will always load, and
+   * show blocks. Our block class records if an instance of it was built when
+   * the page loads. The module registers an event listener to add if the block
+   * was rendered to the response headers.
+   *
+   * We do all this because the page render runs in a different process to this
+   * test, so we can't just read the value from the class directly.
+   *
+   * @dataProvider blockRegionProvider
```
ekes, line 60: OK so this is testing if something is rendered and something isn't rendered. That's pretty much all we need to do here. And as each run is a separate test the static value setting the header will be different and correct.

localgov_base.theme
```diff
@@ -43,22 +43,20 @@ function localgov_base_preprocess_page(&$variables) {
...
   $active_theme = \Drupal::service('theme.manager')->getActiveTheme()->getName();
-  $system_region = system_region_list($active_theme, $show = REGIONS_ALL);
+  $regions = array_keys(system_region_list($active_theme, REGIONS_VISIBLE));
 
-  foreach ($system_region as $key => $value) {
-    $region = $key;
-    $has = 'has_' . $region;
+  $excluded_regions = [
+    'messages',
+    'disabled',
```
ekes, line 50: Does disabled get returned with REGIONS_VISIBLE set on the request? Are there disabled not visible regions? What are they for?

ekes (approved):
I'll say this is makes sense. The test checks as expected.

Would to be good to know about the REGIONS_VISIBLE / 'disabled' thing though.

localgov_base.theme
rupertj: In a theme's info.yml file, there's an optional key called 'regions_hidden', which the docs describe as 'A list of inherited regions to remove.'. If you pass the REGIONS_ALL constant to system_region_list(), these get returned. If you pass REGIONS_VISIBLE, they don't.

So my understanding of this is that it's a mechanism to allow a theme to not inherit a region from its base theme, and we shouldn't try to do anything with regions that haven't been inherited.

PR #500, opened 2023-11-06
adds theme options for unpublished pages

### LINKED ISSUE #495: Display options for unpublished pages
Following on from pink backgrounds in https://github.com/localgovdrupal/localgov_base/pull/476, which haven't been universally popular

We need to provide options which could be:

- turn off pink background for everyone/ specific roles
- don't show pink background on pages accessed by Preview Link
- show the word "Draft" instead of pink background

markconroy:
Closes #495 

Here it is in action: https://youtu.be/yJr1NB8nwlo


finnlewis:
Thanks @markconroy ! 

@ekes  is testing it! 

one quick question, is there a reason that this checks for exactly 0? 
https://github.com/localgovdrupal/localgov_base/pull/500/files#diff-f9772c1a7f4e40b1fa6f9272c087adbfef38038d1d182a377d91d6993ac01255R86

where as in the schema it is defined as a boolean?
https://github.com/localgovdrupal/localgov_base/pull/500/files#diff-12b7cd66524e1d39cf4bcb2cf65bdeb0b9bdcd116ae98b71c06ab195bbedac04R12

Not a biggy, but we're interested!

ekes:
> Thanks @markconroy !
> 
> @ekes is testing it!
> 
> one quick question, is there a reason that this checks for exactly 0? https://github.com/localgovdrupal/localgov_base/pull/500/files#diff-f9772c1a7f4e40b1fa6f9272c087adbfef38038d1d182a377d91d6993ac01255R86
> 
> where as in the schema it is defined as a boolean? https://github.com/localgovdrupal/localgov_base/pull/500/files#diff-12b7cd66524e1d39cf4bcb2cf65bdeb0b9bdcd116ae98b71c06ab195bbedac04R12
> 
> Not a biggy, but we're interested!

Being exact `===` can reveal a lot.

In `localgov_base` it in fact does _not_ work, because the schema is there and it returns a boolean. So the check `theme_get_setting... === 0` has `false === 0` and doesn't equate to true. It does work in `localgov_scarfolk` but that's because of a bug, there's no schema, and so as Drupal doesn't know it's a boolean it returns `0` and `1` which evaluate as 'expected' here.

Now the easy way out of this is to be PHP sloppy and just `==` but we should have a schema, and we should be correct. So we need to add a schema to `localgov_scarfolk` too.

ekes:
> Now the easy way out of this is to be PHP sloppy 

OK So maybe being sloppy might also be a good thing? If others have created their extension of the theme and haven't added a schema it's probably nice if this still works?

In which case we should also change the evaluation from `===` to `==`

markconroy:
> one quick question, is there a reason that this checks for exactly 0?
@finnlewis I was checking for `0` because `1` and `0` were the results that I was getting back from the theme setting when doing `dpm`.

@ekes following your changes, are you happy for us to merge this now?

ekes:
> @ekes following your changes, are you happy for us to merge this now?

Happy assuming someone has tested _my_ changes :) ie it still works, and now for both localgov_base directly and localgov_scarfolk as a derived them.

rupertj:
The idea here is great, but I think the form label for the background option would make more sense the other way round.

EG, Instead of:
"Remove pink background from unpublished content." defaulted to off
It should be:
"Add pink background to unpublished content" defaulted to on.

willguv:
@markconroy could we make @rupertj's change?

markconroy:
@rupertj I'll have to think about that. Will mentioned that to me yesterday as well, but I think that might be a breaking change if site already have the pink set and we are removing it. 

Let me test it tomorrow and see.

markconroy:
@rupertj and @willguv I've added an update so now the question is "Add background colour ..." rather than "Remove background colour ...".

However, this now means that all current live sites will have the colour removed by default unless they explicitly opt-in to having it added back in. Maybe this is desired behaviour.

Any thoughts, or shall we start the merging?
Screenshot 2023-11-09 at 10 32 39

rupertj:
We could add an update hook to set "Add background colour..." to on by default and therefore keep existing behaviour when people upgrade.

I'm happy to do that if you like?

willguv:
That sounds like a good plan @rupertj

localgov_base.theme
```diff
@@ -65,6 +79,46 @@ function localgov_base_preprocess_page(&$variables) {
   if (!isset($variables['localgov_base_remove_css'])) {
     $variables['#attached']['library'][] = 'localgov_base/global';
   }
+
+  // We have a theme setting boolean for removing the pink background from
+  // unpublished content. If that setting is false,
+  // we will add the 'unpublished' library to our page.
+  dpm(theme_get_setting('localgov_base_add_unpublished_background_colour'));
```
rupertj: There's a rogue dpm() here.

localgov_base.theme
```diff
@@ -65,6 +79,46 @@ function localgov_base_preprocess_page(&$variables) {
...
+  if (theme_get_setting('localgov_base_add_unpublished_background_colour') == TRUE) {
```
rupertj, line 86: The '== TRUE' bit of this condition is redundant here (as well as in a few other places in this file).

localgov_base.theme
```diff
@@ -65,6 +79,46 @@ function localgov_base_preprocess_page(&$variables) {
...
+ */
+function localgov_base_preprocess_block(&$variables) {
+  $route_match = Drupal::routeMatch();
+  // Check if we are on a node page.
+  if ($route_match->getRouteName() == 'entity.node.canonical') {
+    $node = $route_match->getParameter('node');
+    $node_status = $node->isPublished();
+    if ($node_status === FALSE) {
+      if (theme_get_setting('localgov_base_add_draft_note_to_unpublished_content') == TRUE) {
+        if ($variables['plugin_id'] === 'localgov_page_header_block') {
```
rupertj: I'd check this first, as it's the most specific condition. It means that in most cases, we'll exit earlier and spend less time checking all the other conditions when we're not looking at this block.

rupertj (commented):
I spotted a few things in the existing code that I'd change when having a quick look for what the update hook needs to do.

PR #503, opened 2023-11-10
Create fields.css

justinepocock:
Fix the bug for inline fields not being inline by stealing this from classy.


finnlewis:
@markconroy is happy!

PR #505, opened 2023-11-13
Swapped the closing and opening p tags.

siliconmeadow:


finnlewis:
Discussing in Merge Monday, @markconroy asked if the <p> tag makes sense, to which the answer is yes! 

It is a plain text field!

PR #507, opened 2023-11-19
changed header searchbox border from none to var(--border)

### LINKED ISSUE #499: Search form box on site not visible in Windows High Contrast Mode
The input box on the search form on all pages of the site has the borders turned off in 'web/themes/contrib/localgov_base/css/components/header.css':

`.lgd-region--search .form-text {
  width: 100%;
  border: var(--header-search-input-border);
}`

This makes the box invisible in Windows HCM.

The box should have a border like the other boxes for "Docs", "Github" and "Apply" alongside it.

`border: 1px solid transparent;`

@msayoung 

high_contrast

bisd:
I changed the setting `--header-search-input-border: none` to `--header-search-input-border: var(--border)` in variables.css. Hope that is ok?

### DIFF
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -181,7 +181,7 @@ body {
   --header-search-button-bg-color: var(--color-accent);
   --header-search-button-color-hover: var(--color-accent);
   --header-search-button-bg-color-hover: var(--color-white);
-  --header-search-input-border: none;
+  --header-search-input-border: var(--border);
 
   /* Secondary Menu */
   --secondary-menu-bg-color-first: var(--color-accent);
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -137,6 +137,7 @@ body {
   --border: var(--border-width) var(--border-style) var(--border-color);
   --border-large: var(--border-width-large) var(--border-style) var(--border-color);
   --border-dark: var(--border-width) var(--border-style) var(--border-color-dark);
+  --border-transparent: 1px solid transparent;
 
   /* Animation */
   --transition-time: 0.3s;
@@ -181,7 +182,7 @@ body {
   --header-search-button-bg-color: var(--color-accent);
   --header-search-button-color-hover: var(--color-accent);
   --header-search-button-bg-color-hover: var(--color-white);
-  --header-search-input-border: var(--border);
+  --header-search-input-border: var(--border-transparent);
 
   /* Secondary Menu */
   --secondary-menu-bg-color-first: var(--color-accent);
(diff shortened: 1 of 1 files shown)


markconroy:
Note: this is a PR to fix the issue in #499

finnlewis:
@markconroy you happy enough with this?

finnlewis:
Discussing in Merge Monday. 

This might place a border on existing websites search form, unless they have specified no border in the child theme.

I will test it and generate some screen shots.

waako:
Transparency is our friend here, rather than having a default colour set from the `--border` variable. 
Maybe introducing a new variable: `--border-transparent: 1px solid`, to be used in other places this crops up.

So we end up with `--header-search-input-border: var(--border-transparent)`.

Header with the 1px transparent border, more or less invisible and shouldn't affect existing sites (where not overriden):
Test site Header, search input has transparent border
What that gives in Windows High Contrast Mode:
Test site Header and windows high contrast mode, search input has visible border

bisd:
I have tried to implement what has been suggested by @waako. The branch has been updated.

css/base/variables.css
```diff
@@ -137,6 +137,7 @@ body {
   --border: var(--border-width) var(--border-style) var(--border-color);
   --border-large: var(--border-width-large) var(--border-style) var(--border-color);
   --border-dark: var(--border-width) var(--border-style) var(--border-color-dark);
+  --border-transparent: 1px solid transparent;
```
markconroy: This should use our default border variables `--border-transparent: var(--border-width) var(--border-style) transparent;`

This is so that if other parts of the item where this is used depend on the border being a certain thickness, we won't be throwing that off by specifying this border as 1px. E.G. if the border is no longer transparent on :hover.

~However, that will put a 1px difference on what we currently have, so a second line in header.css or somewhere to say `border-width: 0;` for this specific item.~

waako: Using existing variables makes total sense, much better than introducing a new one!

Unfortunately adding `border-width: none` would undo what has been added.
So end up back at square one: no visible border on search input with WHCM.

The transparent border is invisible, as it is there from the start it should not affect layout, or not noticeably so.

but some quick responsive testing should hopefully pick out any anomalies.

markconroy: Ok, let's ignore where I asked for `border-width: 0` but keep the initial part where we set the variable to `--border-transparent: var(--border-width) var(--border-style) transparent;`

bisd: Thanks very much @markconroy and @waako the branch has been updated accordingly.

markconroy (approved):

PR #508, opened 2023-11-19
hidden pager items markers as they are visible in WHCM

### LINKED ISSUE #501: Bullit points visible with pager items on multiple pages in Windows High Contrast Mode
The bullit points next to page numbers are visible when Windows High Contrast Mode is turned on. A css quick fix may be to set `[list-style-type]: none;` on the unordered list for the pager items.

An example may be seen on this page:
https://test.localgovdrupal.org/events

This is what it looks like on WHCM:
pager_items

@msayoung

bisd:
I have made the change in pager.css. Hope that is ok?

### DIFF
--- a/css/components/pager.css
+++ b/css/components/pager.css
@@ -3,6 +3,7 @@
   flex-wrap: wrap;
   justify-content: center;
   padding-left: 0;
+  list-style-type: none;
 }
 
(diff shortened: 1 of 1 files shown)


markconroy:
Fixes #501

Issue #513, opened 2023-11-21
[DRAFT] only added to title if page has never been published


AWearring:
Introduced in https://github.com/localgovdrupal/localgov_base/pull/500 the appending of [DRAFT] only seems to happen on newly created pages that haven't been published and not any that are "in draft" after being published.

I think that when viewing the "latest version" or using a "Preview link" that this should also have [Draft] appended to the title

PR #514, opened 2023-11-21
Append [Draft] on latest version and preview link routes

AWearring:
Append [Draft] to title if this page is a unpublished on `latest version` and `preview link` routes, as well as on node pages

See issue for more details: https://github.com/localgovdrupal/localgov_base/issues/513

This PR adds additional route conditions `entity.node.latest_version` and `entity.node.preview_link`

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -102,13 +102,13 @@ function localgov_base_preprocess_node(&$variables) {
  * Implements hook_preprocess_HOOK().
  */
 function localgov_base_preprocess_block(&$variables) {
-  $route_match = Drupal::routeMatch();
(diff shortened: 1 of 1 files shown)


AWearring:
@tran-bigbluedoor has suggested that It is better if we can move the `Plugin_id` condition to the top level to avoid getting all the `route_match` and `node` for every single block all the site.

I'll update the PR accordingly

AWearring:
I have had some additional feedback from another senior developer at Big Blue Door suggesting that we should be testing $node is a "node entity" as they have had situations where some routes can return just node id instead of full entity object

markconroy (approved):

markconroy:
Thanks @AWearring. Looks good to me.

PR #516, opened 2023-11-27
updates page-sections to use a scoped variable approach

### LINKED ISSUE #515: Set page section colours via scoped variables
At the moment, for our `.lgd-page-section--*` background colours and text colours, we are defining them via global CSS variables. This works for the most part, but will mean maintaining a lot of code if we have `::before` and `::after` type things and we'd like the colours to `inherit`.

for example, if a page section has a shape at the top of it, we want to be able to do something like

```
.lgd-page-section::after {
  background-color: inherit;
}
```
In our current set up, we'd need to do this:

```
.lgd-page-section--background-color-1 {
  background-color: var(--color-accent);
}
.lgd-page-section--background-color-2 {
  background-color: var(--color-accent-alt);
}
.lgd-page-section--background-color-3 {
  background-color: var(--color-theme-1);
}
.lgd-page-section--background-color-4 {
  background-color: var(--color-theme-2);
}
.lgd-page-section--background-color-5 {
  background-color: var(--color-red);
}
.lgd-page-section--background-color-6 {
  background-color: var(--color-pink);
}
.lgd-page-section--background-color-1::after {
  background-color: var(--color-accent);
}
.lgd-page-section--background-color-2::after {
  background-color: var(--color-accent-alt);
}
.lgd-page-section--background-color-3::after {
  background-color: var(--color-theme-1);
}
.lgd-page-section--background-color-4::after {
  background-color: var(--color-theme-2);
}
.lgd-page-section--background-color-5::after {
  background-color: var(--color-red);
}
.lgd-page-section--background-color-6::after {
  background-color: var(--color-pink);
}
```

If we scope the CSS variables, we can then set a default variable for the background colour, and re-set that for each variation, then let that inherit where we want it. This will be particularly powerful in sub-themes.

In a subtheme, we can simple do something like this:

```
.lgd-page-section {
  --color-page-section-background-color-1: red;
  --color-page-section-background-color-2: green;
  --color-page-section-background-color-3: blue;
  --color-page-section-background-color-4: pink;
  --color-page-section-background-color-5: brown;
}
.lgd-page-section::before,
.lgd-page-section::after {
  background-color: inherit;
}
```

markconroy:
Closes #515

### DIFF
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -445,6 +445,8 @@ body {
   --blog-author-content-margin: var(--spacing-largest);
 
   /* Page Sections */
+  --color-page-section-background-color: transparent;
+  --color-page-section-text-color: var(--color-black);
   --color-page-section-background-color-1: var(--color-accent);
   --color-page-section-background-color-1-text-color: var(--color-white);
   --color-page-section-background-color-1-link-color: var(--color-white);
--- a/css/components/page-sections.css
+++ b/css/components/page-sections.css
@@ -1,30 +1,34 @@
+.lgd-page-section {
+  background-color: var(--color-page-section-background-color);
+  color: var(--color-page-section-text-color);
+}
 .lgd-page-section--bg-colour-1 {
-  color: var(--color-page-section-background-color-1-text-color);
-  background-color: var(--color-page-section-background-color-1);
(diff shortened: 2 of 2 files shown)


kiwimind (approved):
This looks good. Fairly trivial code change with definite improvement to use cases going forwards.

PR #519, opened 2023-12-19
sets tabs to be sticky when logged in

### LINKED ISSUE #518: Make the tabs region sticky when logged in
Some content designers have mentioned that the view/edit/delete etc buttons should always be visible, and that is it annoyed on a very long page to have to keep scrolling up to the top to find them when you want to edit something.

Let's make them sticky for logged in users, then no matter how far down the page you are they will be within easy reach.

Screenshot 2023-12-19 at 08 42 44

markconroy:
Closes #518

### DIFF
--- a/css/components/region-tabs.css
+++ b/css/components/region-tabs.css
@@ -46,3 +46,13 @@
   position: relative; /* So the outline doesn't hide behind the next item */
   outline: 2px dotted;
 }
+
+/*
+  Make the tabs region sticky for editors, so they do not have to
+  scroll to the top of the page to find the edit link.
+*/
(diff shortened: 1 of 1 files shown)


markconroy:
@finnlewis @msayoung any chance of a review of this? We have it live on Essex Co Co already and working fine.

finnlewis:
Tested locally and works fine in Firefox and Chrome on Ubuntu. 

Looks like the tests have just all passed too, since merging back from 1.x with @stephen-cox's recent fixes.

finnlewis (approved):

PR #522, opened 2024-01-26
closes the secondary menu if you click outside it

### LINKED ISSUE #517: Services header drop down menu not closing
As a user I want the services menu (in header) to disappear/contract if I click anywhere that isn't a link in the services drop down menu in the header.

If you click anywhere on the page that isn't in the menu then then menu should go away.

markconroy:
Closes #517

### DIFF
--- a/js/header.js
+++ b/js/header.js
@@ -18,6 +18,7 @@
       // Set up initial variables.
       //
       // We need a bunch of classes and selectors.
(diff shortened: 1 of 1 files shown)


finnlewis (approved):
Tested locally and definitely works for me. 

Thanks @markconroy  and @davidupjohn

PR #530, opened 2024-03-15
checks for heading before printing it

### LINKED ISSUE #529: The heading tag is printing whether there is content or not.
https://github.com/localgovdrupal/localgov_base/blob/19709031897cd70eae6f3f88b2be2f15268e04a6/templates/paragraphs/paragraph--localgov-call-out-box.html.twig#L63

markconroy:
Closes #529


tqqa
justinepocock, line 1: uh, what's this file?

templates/paragraphs/paragraph--localgov-call-out-box.html.twig
justinepocock, line 1: This bit is golden!

justinepocock (commented):
I don't know what the second tqqa file is but the actual check is fine.

justinepocock (approved):

PR #537, opened 2024-04-05
adds new grid system for use in views

### LINKED ISSUE #536: Add a grid system for views
it would be handy to be able to drop css classes into our views for such as `view--grid-thirds` and let that automatically layout items in one column on mobile, 2 columns on tablet, and 3 columns on large screens.

Classes to include:
  - view--grid-halves
  - view--grid-thirds
  - view--grid-quarters
  - view--grid-fifths

  As well as that, can can specify the gap with with the following classes:
  - view--grid-gap-none
  - view--grid-gap-small
  - view--grid-gap-smaller
  - view--grid-gap-smallest
  - view--grid-gap-medium
  - view--grid-gap-large
  - view--grid-gap-larger
  - view--grid-gap-largest

We _might_ be able to extend our current `lgd-row` type system for this purpose. 

While this will not suffice every layout for every view, I reckon it will suffice for 90% of them. Then we can easily drop the classes into the views CSS field.

Screenshot 2024-04-04 at 16 30 57

markconroy:
closes #536 

This is a very simple to use grid system to allow us add classes to views UI, such as `view--grid` or `view--grid--thirds`.

Once merged, we can then look at adding a more generic class to it such as `lgd-grid` which can be used instead of our `lgd-row` classes if we wish.

### DIFF
--- a/css/layout/grid.css
+++ b/css/layout/grid.css
@@ -79,3 +79,80 @@
     width: calc(75% - var(--grid-column-spacing));
   }
 }
+
+/*
+  Any view using the `view--grid` class will have:
+  - 1 column on mobile
+  - 2 columns on tablet and desktop
+  - 3, 4, 5 columns on desktop depending on what extra class you add to the view.
+  - a gap of var(--spacing) between each item, unless you add a gap class.
+
+  Available classes:
+  - view--grid-thirds
+  - view--grid-quarters
+  - view--grid-fifths
+
+  As well as that, can can specify the gap with with the following classes:
+  - view--grid-gap-none
+  - view--grid-gap-small
+  - view--grid-gap-smaller
+  - view--grid-gap-smallest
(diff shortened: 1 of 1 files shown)


css/layout/grid.css
```diff
@@ -79,3 +79,80 @@
...
+
+.view--grid .view-content {
+  display: grid;
+  grid-template-columns: repeat(var(--ecc-view-grid-columns), 1fr);
+  grid-gap: var(--ecc-view-grid-gap);
+}
+
+@media screen and (min-width: 48rem) {
+  .view--grid .view-content {
+    --ecc-view-grid-columns: 2;
```
justinepocock: Are all these -ecc--'s for Essex?

finnlewis:
Discussing in Merge Tuesday, leaving this for @msayoung / @justinepocock to review and comment / approve.
Looks good to me!

justinepocock (approved):

PR #542, opened 2024-05-02
Added subsites_menu region.

rupertj:


dedavidson (dismissed):
I feel the following commits would be better in the [subsite extras](https://github.com/localgovdrupal/localgov_subsites_extras) module as documentation/examples rather than being merged into the base theme:

- f0e9bc121b41fda657a4f943af3fc3eaee9e87c9
- c0d87dc1caa7be6bcccb9659fba95826b143a5ac
- e605277fcfb5932325d255d560d4e286509f726d

justinepocock:
I was in two minds about that myself.

markconroy:
Hi,

I'd prefer the HTML and CSS for this to be in this theme rather than in the module. We can add HTML and CSS to the module as well if we want so that it works even if `localgov_base` is not the base theme being used.

Keep the variables here means we can place them all in the variables.css file for consistency, and it will make it easier for sub-themes to find them and override them.

dedavidson:
Hi @markconroy 
The HTML and CSS in the three commits are examples though and don't do anything unless a menu called 'subsites' is created. They will have to be overridden/re-created for whatever the specific implementation calls its menus.

markconroy:
Hi @dedavidson, 

Yep, I've noticed that. I'm doing a rewrite of this PR at the moment, hoping to have something ready for it by EOD today, if not, I'll definitely have it ready over the weekend.

rupertj:
> The HTML and CSS in the three commits are examples though and don't do anything unless a menu called 'subsites' is created. They will have to be overridden/re-created for whatever the specific implementation calls its menus.

Subsites extras does create a menu called 'subsites' by default on install now.

markconroy:
I've a pretty major re-write of this done now if anyone wants to review it. 

I've no JS done yet. Let's get this reviewed and merged and I'll get the JS done then.

Also, related PR: https://github.com/localgovdrupal/localgov_scarfolk/pull/36 (it's just one line of code)

markconroy:
A few small updates since Friday, and the JS added for the menu to toggle on/off on small screens.

justinepocock (approved):
cheers mark

Issue #535, opened 2024-04-03
Status Pages won't display overridden content


dedavidson:
If status pages are overridden in for example microsites by specific nodes then the templates for `page--404.html.twig` and `page--403.html.twig` will not display the content of these nodes as their `page_content` block overrides the same block from `page-status.html.twig`.

chriswales95:
Okay, did a deep dive and found a few issues and will open up a MR here and potentially another in localgov_core.

The 403 template, page--403.html.twig, doesn't actually do anything. It extends @localgov_base/layout/page.html.twig but that template doesn't have a block called page_content. That means dynamic content will never load.

The base theme templates hard-code the text in them which makes it difficult to customise the contents. I'd imagine users would like full control over what's there. For example, on the 404 you get dynamic content and the hard-coded text. Drupal's default page content for these pages is quite sparse and unhelpful which is why I imagine the content has been put in these templates. 

It's not possible to check if the site has a custom 404 or 403 in the template itself. The `page.content` variable always has something in it even if it's just Drupal's default content. We can check this by inspecting the sysem.site config but this doens't work as well on Microsites in which each microsite can have it's own system.site config. 

Drupal checks for the presence of custom error pages in the event subscriber CustomPageExceptionHtmlSubscriber but the actual content is set in the controller Http4xxController. I sugest we move the hard-coded text into another controller (perhaps we override the Http4xxController in localgov_core).

So my suggestions are:
1. We remove the hard-coded content from the templates and have them simply extend @localgov_base/layout/status-pages/page-status.html.twig
2. In another issue / MR, Remove the hard-coded text and override Http4xxController and set it there.


I will open a PR for item 1 here and open a seperate issue in localgov_core for point 2.

PR #544, opened 2024-05-22
Fix status messages and dynamic content

chriswales95:
Removes the hard-coded messages in the 403 and 404 pages.

Also changes the 403 page to use the page status template to allow dynamic content.

This relates to issue #535.

I'm going to open an issue and pull request in localgov_core for the other issue identified in that issue.


markconroy (approved):
Thanks for this @chriswales95

PR #546, opened 2024-05-27
adds add-to-calendar functionality for events

### LINKED ISSUE #532: 'Add to calendar' for Events
Almost done - try it at

http://essex-1.mc.anner.ie/events/single-event
http://essex-1.mc.anner.ie/events/test-repeating-event

markconroy:
Closes #532 

===


templates/_components/add-to-calendar.twig
```diff
@@ -0,0 +1,72 @@
...
+      &location={{ atc_location }}
+    {% endif %}
+    {% if atc_rrule %}
+      &recur=RRULE:{{ atc_rrule }}
+    {% endif %}
+{% endset %}
+
+<div{{ lgdAddToCalendarAttributes.addClass(classes) }}>
+
+<button class="localgov-add-to-calendar__trigger" type="button" name="addToCalendarTrigger" aria-label="{{ 'Toggle List of Calendars to Add Event To'|t }}">
```
msayoung: aria-label isn't required

templates/_components/add-to-calendar.twig
```diff
@@ -0,0 +1,72 @@
...
+    }
+  %}
+  <span class="localgov-add-to-calendar__trigger-text">
+    {{ 'Add to Calendar'|t }}
+  </span>
+</button>
+
+  <dialog class="localgov-add-to-calendar__dialog">
+    <div class="localgov-add-to-calendar__dialog-close">
+      <button class="localgov-add-to-calendar__dialog-close-button" type="button" name="closeAddToCalendarDialog" aria-label="{{ 'Close Add to Calendar Dialog'|t }}">
```
msayoung: Again, aria-label isn't needed IMO

templates/_components/add-to-calendar.twig
```diff
@@ -0,0 +1,72 @@
...
+        {{ 'Close'|t }}
+      </button>
+    </div>
+    <div class="localgov-add-to-calendar__items">
+
+      <div class="localgov-add-to-calendar__item">
+        <a href="{{ google_link }}" target="_blank" rel="nofollow">{{ 'Add to Google Calendar'|t }}</a>
```
msayoung, line 58: I think these should be Buttons rather than links

msayoung:
I notice testing on FF  and Chrome  with NVDA screenreader that when the dialog opens the entire thing is read out. I think because focus is moving to the dialog rather than a component within it. But it may just be an issue with NVDA. 

https://adrianroselli.com/2020/10/dialog-focus-in-screen-readers.html

The dialog probably wants a name of some sort. A heading, or an aria-label or aria-labelledby would be good.

templates/_components/add-to-calendar.twig
```diff
@@ -0,0 +1,72 @@
...
+set classes = [
+  "localgov-add-to-calendar"
+]
+%}
+
+{{ attach_library('localgov_base/add-to-calendar') }}
+
+{% set lgdAddToCalendarAttributes = create_attribute() %}
+
+{% set ical_link %}data:text/calendar;charset=utf8,BEGIN:VCALENDAR%0AVERSION:2.0%0ABEGIN:VEVENT{% if atc_start_date %}%0ADTSTART:{{ atc_start_date|date("Ymd\\THi00\\Z") }}{% endif %}{% if atc_end_date %}%0ADTEND:{{ atc_end_date|date("Ymd\\THi00\\Z") }}{% endif %}{% if atc_title %}%0ASUMMARY:{{ atc_title }}{% endif %}{% if atc_details %}%0ADESCRIPTION: {{ atc_details|replace({'<br />': '\n', '<br>': '\\n', '<p>': '', '</p>': '\\n\\n'}) }}{% endif %}{% if atc_location %}%0ALOCATION: {{ atc_location|replace({'<br />': ' ', '<br>': ' ', '<p>': ' ', '</p>': ''})}}{% endif %}{% if atc_rrule %} %0ARRULE: {{ atc_rrule }} {% endif %}%0AEND:VEVENT%0AEND:VCALENDAR{% endset %}
```
ekes, line 11: ical uses CRLF line endings so it should be `\r\n` instead of `\n` but then that doesn't mean the same thing as a line break. So text might will end up in a single line. Also and we certainly can ignore this, but line lengths should be 75 characters.

What is going to be more patchy for client support is having HTML in the description. Unless this has already been converted to plain text? Any text including HTML should be an additional version using ALTREP https://www.rfc-editor.org/rfc/rfc5545#section-3.2.1

Not important either but should probably output the date including the seconds, just in case someone (WHY?) it setting the seconds. Ymd\THis

Other comments would be enhancements:
 - We can add the geo location point `GEO:37.386013;-122.082932` which is lat;long
 - Doing it with timezone rather than Z would be nice, but lots of work.

markconroy:
@ekes 
> ical uses CRLF line endings so it should be \r\n instead of \n but then that doesn't mean the same thing as a line break. So text might will end up in a single line.

Any preference on what we do here? `\r\n` or `\n`? I guess we want to be right, so `\r\n` but we also don't want everything on one line, so `\n`. My preference would be to leave it as is if it's working and figure out something better if a problem presents itself.

> Any text including HTML should be an additional version using ALTREP

I'm not fully sure exactly what we need to do here? I don't like having the body field used as the event description. So perhaps it's best to just remove that section? I can imagine an event with a very long body field being not exactly what a user is expecting in their calendar. Maybe we should just use the summary field (which is just plain text with no HTML) as our value here instead?

templates/_components/add-to-calendar.twig
```diff
@@ -0,0 +1,72 @@
...
+
+{% set google_link %}
+    https://www.google.com/calendar/r/eventedit?text={{ atc_title }}
```
joachim-n, line 14: This assembly of URLs would be better done in PHP code rather than in twig.

I think a more flexible solution would be to build this as a formatter alterer for the date field.

joachim-n: It looks like this contrib module does what I mean - https://www.drupal.org/project/addtocal_augment. Though I don't know what SmartDate is -- it looks like a different ecosystem from Date Recur module that we use, so probably not suitable for us. 

The same approach would work though -- although it should be simpler than that one, as I don't see what Date Augmenter API is needed, since Drupal core provides the means to alter field formatters.

markconroy: > This assembly of URLs would be better done in PHP code rather than in twig.

Do you want to amend the PR to do this?

finnlewis: @joachim-n thanks for your comments, just discussing in Merge Tuesday, are you planning to do some work on this?

markconroy:
Should we merge this? The only thing outstanding is whether to use twig or PHP to construct our URLs. This seems more like an enhancement than a blocker to me.

templates/_components/add-to-calendar.twig
ekes: So having stared at this for a bit.

As it stands, realistically, just moving the URL assembly to PHP would be moving the construction of some of the output from the twig template to the preprocess. 

What joachim also points to would be the more ideal of having this add to the display widget itself, the additional bonus there would be that it can be enabled / disabled. But there isn't really a convenient way of adding to the Formatter Plugin itself (what would it be a hook_form_alter, and also a template override?). Making the formatter pluggable itself to add features would be a huge bonus that could be contributed back but feels like a few hours work?

ekes: I will emphasis - as it stands - as there is no config to switch this on or off it will appear on every site inheriting the base theme, and to remove it (no I don't know why people would want to do that) requires extending the the template to remove what is added to the event full template.

markconroy: We could add a new boolean field to the event content type to say "add Add to Calendar" widget to this event" and have it switched on by default for all new events. I'm not sure if that's overkill, or if that a "next phase of work" job for us.

markconroy: @ekes how about we merge this as is, then later

1. If someone wants to move the construction of the URL from twig to PHP they can do that in a follow up issue.
2. If we think it's a good idea to have a checkbox on each event to "Turn off add to calendar for this event" we can create a follow-up issue for this (in the localgov_events repo)

js/add-to-calendar.js
```diff
@@ -0,0 +1,23 @@
...
+      const addToCalendarButtons = once('all-add-to-calendar-buttons', '.localgov-add-to-calendar__trigger', context);
+      if (addToCalendarButtons.length) {
+        addToCalendarButtons.forEach(button => {
+          const calendarModal = button.nextElementSibling;
+          const calendarModalClose = calendarModal.querySelector('.localgov-add-to-calendar__dialog-close-button');
+          button.addEventListener('click', () => {
+            calendarModal.showModal();
+          });
+          calendarModalClose.addEventListener('click', () => {
+            calendarModal.close();
```
anthonylindsay, line 17: If we're launching a modal we'll need to pay attention to two things: focus and close actions.
We'll need to ensure that focus is in a sensible place in the modal, and then on close, that it goes back to a sensible place. Maybe back to the button that launched it?

For closing, we need to make sure that a modal can be closed by

- clicking a close button (which must be accessible by the keyboard, and might make a good sensible focus starting point?)
- clicking outside the modal
- hitting `esc`

And usually we'll want to do something to prevent page scroll whilst the modal is open, coz it's just weird.
Has LGD not already got any modal constructions in any of the other myriad modules we can lean on, which might have these things already addressed?

templates/_components/add-to-calendar.twig
ekes: Yep fine with me.

finnlewis:
Discussing this in Merge Tuesday: 

Enabling by default on existing sites, for sites that have not overridden this template, will automatically add the button to 'Add to calendar' to event node full view.

We're happy with this (@willguv, @markconroy,  @ekes, @stephen-cox, @FinnERLewis )

@andybroomfield would we consider moving this to the events module, to a pseudo field? This can be a follow up issue.

PR #549, opened 2024-05-28
Add support for layout paragraphs builder on front end

### LINKED ISSUE #548: Add support for layout paragraphs builder field formatter in front end
Noting there is also a new experimental field formatter called layout paragraphs builder.
This seems to be what Mercury editor is using behind the scenes.
To work with Subsites, localgov_base needs to be altered to allow for it.
Changing 
```twig
  {% if node.localgov_subsites_content.value %}
    {{ content.localgov_subsites_content }}
  {% endif %}
```
to
```twig
  {% if node.localgov_subsites_content.value or content.localgov_subsites_content['#type'] == 'layout_paragraphs_builder' %}
    {{ content.localgov_subsites_content }}
  {% endif %}
```

_Originally posted by @andybroomfield in https://github.com/localgovdrupal/localgov/issues/711#issuecomment-2120226369_

andybroomfield:
Fix #548

Allows the subsites content field formatter to be changed to layout paragraphs builder which then allows subsites to be edited in the front end, even if there are no paragraphs present. This is important if using the front end theme for content editing (eg. Using mercury editor).

### DIFF
--- a/templates/content/node--full.html.twig
+++ b/templates/content/node--full.html.twig
@@ -148,7 +148,7 @@
       </div>
   </div>
 
-  {% if node.localgov_subsites_content.value %}
+  {% if node.localgov_subsites_content.value or content.localgov_subsites_content['#type'] == 'layout_paragraphs_builder' %}
     {{ content.localgov_subsites_content }}
   {% endif %}
   


andybroomfield:
As discussed on https://github.com/localgovdrupal/localgov/issues/711.

This is to allow front end page builders, with this fix:
Screenshot 2024-05-28 at 6 35 21 PM

markconroy (approved):
LGTM

PR #558, opened 2024-06-06
feat: makes core/drupal a global dependency

### LINKED ISSUE #557: core/drupal should be a global dependency
I'm working on an LGD implementation that can't use the `localgov_base` [`header`](https://github.com/localgovdrupal/localgov_base/blob/1.x/localgov_base.libraries.yml#L62-L68) library, so I excluded it with the following in the sub-theme's info file:

```yml
libraries-override:
  localgov_base/header-js: false
```
Immediately after doing this, all javascript in the sub-theme started to fail with `Uncaught ReferenceError: Drupal is not defined` in the browser console.

I was able to correct the problem in the sub-theme by adding `core/drupal` to the sub-theme's `global` library.

Since *all* Drupal behaviours (including [`localgov_base/guides`](https://github.com/localgovdrupal/localgov_base/blob/1.x/localgov_base.libraries.yml#L135-L137)) rely on `core/drupal`, and since sub-themes may replace some/all libraries, IMO `core/drupal` should probably become a global library.

ctorgalson:
This helps not to break overrides by subthemes that override e.g. localgov_base/header-js without making core/drupal a requirement.

Closes #557

### DIFF
--- a/localgov_base.libraries.yml
+++ b/localgov_base.libraries.yml
@@ -17,6 +17,7 @@ global:
   # Declare dependencies for other libraries that we want to load
   # on every page.
   dependencies:
+    - core/drupal
     - localgov_base/grid
     - localgov_base/header
     - localgov_base/footer
@@ -63,7 +64,6 @@ header-js:
   js:
     js/header.js: {}
   dependencies:
-    - core/drupal
     - core/drupal.debounce
     - core/once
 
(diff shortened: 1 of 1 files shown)


markconroy (approved):
Looks good to me. Thanks @ctorgalson

PR #559, opened 2024-06-07
Add generic prev/next template for blogs/guides/etc

### LINKED ISSUE #545: `localgov_blogs` navigation template
Add the `localgov_blogs` navigation template to `localgov_base`

markconroy:
Closes #545 

We have a nice pattern for our prev/next block for Guides. Let's make that a little bit more generic and then use the same pattern for the prev-next block for blogs as well.


willguv:
Nice one @markconroy

AWearring:
I like the approach - if possible do you think it would be worth including the Publications and Step-by-step Next/Prev too

markconroy:
Publications = definitely
Step by step = a little trickier as it's a views block rather than a custom block.

If we like the approach, let's get this approved and merged and create follow-up issues for the other content types

PR #561, opened 2024-06-11
Adds subtitle variable for header block

### LINKED ISSUE #560: Create variable for subtitle for page title/header block
LocalGov Core now has a `{{ subtitle }}` variable. Let's add the same to `localgov_base`.

Note: this will do nothing until a theme implements an event subscriber to set a field as the subtitle field. We are only putting the variable in position so we have it for themes that might use it.

Further details:
- https://github.com/localgovdrupal/localgov_core/releases/tag/2.13.0
- https://github.com/localgovdrupal/localgov_guides/issues/122

===

markconroy:
Closes #560.

### DIFF
--- a/templates/block/localgov-page-header-block.html.twig
+++ b/templates/block/localgov-page-header-block.html.twig
@@ -26,7 +26,7 @@
       <h1 class="lgd-page-title-block__title">
         {{ title }}
         {% if subtitle %}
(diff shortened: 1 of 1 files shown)


finnlewis (approved):
Sounds good, thanks @markconroy !

PR #562, opened 2024-06-12
Adds icons for callouts and alerts

### LINKED ISSUE #531: Add alert icon, like exclamation mark in black circle, to call out and alert styles
At Haringey we've noticed that some people ignore blue call outs, where we put info we want to draw to their attention, and skip over them to get to the next section of text. We think people may also be ignoring yellow and red alert styling, although we want to test this.

Is it possible for callout and alert styling to do something like GOV.UK do which is add a black circle with an exclamation mark to the callout/alert? We've noticed this is ignored less. We then aren't just relying on colour to draw someone's attention to something important.

HaringeyBlueCallout
GOVUKAlert

markconroy:
Closes #531 

Here's a PR to add an icon to alerts and callouts. Just to note a few things:

1. This is a change that councils may not be expecting
2. Are we sure callouts AND alerts should have an icons, not just alerts?
3. Some councils may have already implemented a similar fix like this already, so this might affect what they have done.

Let's make sure we have these as notes in the release notes. Here's a screenshot.

Screenshot 2024-06-12 at 16 37 05

===

### DIFF
--- a/css/components/callout.css
+++ b/css/components/callout.css
@@ -2,11 +2,27 @@
   @file Base styles for call out paragraphs / advanced callout div container
 */
 .callout {
+  display: flex;
+  gap: var(--spacing);
   padding: 1.5rem;
-  text-align: center;
+  /* text-align: center; */
   color: var(--color-white);
   font-size: var(--font-size-large);
   font-weight: bold;
+}
+
+.callout::before {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  width: 5rem;
+  height: 2.5rem;
+  padding: var(--spacing-small);
+  content: "!";
+  color: var(--color-black);
+  border-radius: 999rem;
+  background-color: var(--color-white);
+  font-size: 1.5rem;
 }
 
 .callout a {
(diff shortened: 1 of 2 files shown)


msayoung:
> Are we sure callouts AND alerts should have an icons, not just alerts?

There's an argument to add them to the ones named callout danger, and callout success - however the others could be used in all sorts of ways, so I guess maybe not. 

Other thoughts: 

In the gov.uk pattern https://design-system.service.gov.uk/components/warning-text/
there is visually-hidden text "Warning" which will convey a similar message to non-visual users. 

As we're just working with the css, we could potentially add "Warning" as content in the ::before 

However, is "warning" always appropriate? 
Perhaps I'm overthinking this. 


Related: https://github.com/localgovdrupal/localgov_base/issues/563

markconroy:
Given that these are wysiwyg styles, it's tricky to add the visually hidden text. I guess we could do it with JS. This might be more a feature for the alerts than the callouts though, as they are "alerts" rather than "here's a bit of info we want to make look different".

anthonylindsay:
Presumably, given the intent is to convey information here, the icon should be paired to the colour/state?
So an 'info' alert might have one icon, and a 'panic' alert might have another?
Otherwise we're back to using colour to convey meaning without a supplemental method.

PR #567, opened 2024-06-16
fix: Drupal 11 support

millnut:
- Drops Drupal 8 and 9 support
- Sets the minimum Drupal 10 to 10.1.3 so DeprecationHelper can be used (https://www.drupal.org/node/3379306)
- Added Drupal 11 support

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -6,6 +6,7 @@
  */
 
 use Drupal\Component\Utility\Crypt;
+use Drupal\Component\Utility\DeprecationHelper;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\views\ViewExecutable;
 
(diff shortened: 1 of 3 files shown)


localgov_base.theme
```diff
@@ -69,7 +70,16 @@ function localgov_base_preprocess_page(&$variables) {
...
+
+    /** @var \Drupal\Core\Render\RendererInterface $renderer */
+    $renderer = \Drupal::service('renderer');
+    $rendered = DeprecationHelper::backwardsCompatibleCall(
+      currentVersion: \Drupal::VERSION,
+      deprecatedVersion: '10.3',
+      currentCallable: fn() => $renderer->renderInIsolation($copy),
+      deprecatedCallable: fn() => $renderer->renderPlain($copy),
+    );
+
```
markconroy, line 82: I'm guessing this is something we need to fix for D11 rather than just marking it as deprecated?

If so, should we try fix it now, or create a follow up for it?

millnut: Correct, this allows support for both D10 and D11. renderInIsolation was added in Drupal 10.3 so the deprecation helper allows us to support <10.3 and >=10.3 without requiring multiple versions of the module. 

When the deprecation helper is used it does not flag the deprecation as it is wrapped in the helper, however when we decide to no longer support <10.3 (Drupal 10 is supported until mid/late 2026) we can then remove the helper completely. We should create a ticket for removing this helper when we decide to drop support for <10.3

markconroy: Great. Let's get this merged so and that follow up created. Thanks Lee.

===

markconroy (approved):
LGTM

PR #581, opened 2024-08-06
Fix/580/alert banner dialog

### LINKED ISSUE #580: Full page alert banner does not cover the full page
Maybe I'm wrong in what I am expecting, but I thought the full page alert banner would cover the full screen and not let me see the rest of the page until I dismissed it.

What I am seeing in an alert banner placed inside a `dialog` element, which does take over the page but we can still see everything behind it.

I think we need - at minimum - to add a darker `::backdrop` to the alert-banner dialog so the rest of the screen is not so visible.

I think we should also adjust the `margin` so it's set to `auto` for this item, so it's placed in the middle of the screen vertically and horizontally.

Here's what I am currently seeing:

Screenshot 2024-08-05 at 14 55 35

---

markconroy:
Closes #580 

---
See notes: https://github.com/localgovdrupal/localgov_alert_banner/pull/360

---


andybroomfield (approved):
Working for me, thanks @markconroy 

One thing to note is the [change in ::backdrop behaviour](https://developer.chrome.com/blog/css-backdrop-inheritance), which means in pre Chrome 122 it does not inherit the custom properties declared in variables.css so I see no background colour (fine in Firefox and edge and assume my Chomre when it updates).

Chrome 120:
Screenshot 2024-08-06 at 1 08 36 PM

Firefox:
Screenshot 2024-08-06 at 1 08 19 PM

Issue #584, opened 2024-08-13
Add the ability to print a token in the breadcrumb wrappers


rupertj:
I'm fixing up the breadcrumb trail on H&F's moderngov integration. It currently doesn't look great, as the moderngov module uses its own markup, which doesn't match the markup that localgov_base provides. I'd like to fix that by using the real breadcrumb template, but need to be able to insert a token inside the list that moderngov'll replace with the breadcrumbs.

EG, this is what I currently have in H&F's (updated) moderngov template:

```
      {# Copied from breadcrumb.html.twig and modified #}
      {{ attach_library('localgov_base/breadcrumbs') }}
      <nav class="breadcrumbs" aria-labelledby="system-breadcrumbs">
        <h2 id="system-breadcrumbs" class="visually-hidden">{{ 'Breadcrumbs'|t }}</h2>
        <ol class="breadcrumbs__list">
            <li class="breadcrumbs__item">
                <a class="breadcrumbs__link" href="{{ url('<front>') }}">{{ 'Home'|t}}</a>
            </li>
            {breadcrumb}
        </ol>
      </nav>
      {# End copied from breadcrumb.html.twig #}
```

If I can get that "{breadcrumb}" token in there in the right place, I can use the real template instead of copying stuff in, plus it'll fix this for everyone else :)

PR #585, opened 2024-08-13
Adds a variable to let us print a token for replacement in the breadcrumb trail.

rupertj:
Implements #584

### DIFF
--- a/templates/navigation/breadcrumb.html.twig
+++ b/templates/navigation/breadcrumb.html.twig
@@ -5,6 +5,7 @@
  *
  * Available variables:
  * - breadcrumb: Breadcrumb trail items.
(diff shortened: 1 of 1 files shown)


finnlewis (approved):
Looks good to me @markconroy and tests pass.

markconroy:
@rupertj do we need two `{{` around the variable for ModernGov to use or just  one `{`?

rupertj:
It'll be one in the resulting output, but we're printing the value of a variable here, so it's two.

For the purposes of making Modern.Gov work, we'll pass in {breadcrumb} as the value of the var, but if any other reverse CMS thing wants to use it (Maybe Netcall?) they can pass in what they need instead.

markconroy:
Sounds great, let's get this merged.

Issue #564, opened 2024-06-13
Use prev-next.twig for publications prev-next


markconroy:

ctorgalson:
@markconroy Questions about this one: is the intention for `localgov_base` to change the _look_ of the Publications prev/next nav along with the markup (see screenshot)?

![image](https://github.com/user-attachments/assets/8e38589c-f119-4328-9072-bc763ea254eb)
 (Left: current, right: using `localgov_base`'s `prev-next.twig`)
 
 If the look and feel are to stay as-is, do we want to address it in CSS, or should we add condition in `prev-next.twig` that leaves out the default classes if, say, a variable like `omit_template_classes` is defined?

markconroy:
The intention is to have all Prev/Next components looking identical, so if any council changes their design for prev/next they will only need to update the CSS in one place.

ctorgalson:
@markconroy I didn't actually push up any code for this yet :)

Does the design uniformity extend to the positioning? On Publications, the Prev/Next buttons are in the main content area, but there's a sidebar, so they're not at the far left of the layout.

markconroy:
Sorry, I don't even remember clicking the close button. 

Yes, I think it should also extend to the positioning. If that's more difficult, maybe a follow-up issue will be good enough and we just focus on the style/template in this issue.

ctorgalson:
@markconroy created #596. 

It looks ok, but moving the nav to a full-width region probably means creating a block for the nav as is done for other components (the current nav is part of the node content, so can't really move out of the region including the sidebar).

ctorgalson:
Closed by #596

PR #588, opened 2024-08-21
feat: replaces standalone prev-next code with included version

ctorgalson:
## What does this change?

The template `views-view-list--localgov-step-by-step-navigation--prev-next.html.twig` now uses the `_components/prev-next.twig` template, bringing it in line with guides and the localgov-blogs block template.

This has several side-effects, all of them small, one of them (I think) positive:

- the aria-label attribute is lost
- prev/next links now wrapped in div.lgd-prev_next__label
- labels now say "Previous" and "Next" instead of "Previous Step" and "Next Step"
- labels are now bold
- distance between links and footer now *more* consistent with other CTs

## How to test

Compare e.g. `/adult-health-and-social-care/step-by-step/request-support-adult-step-step/apply-financial-support` on an instance running this branch with [the current demo site](https://demo.localgovdrupal.org/adult-health-and-social-care/step-by-step/request-support-adult-step-step/apply-financial-support)

This makes the markup and CSS more consistent with other identical UI components.

There are two questions about the effects:

1. is/was the aria-label on _this_ specific component deliberate and/or important?
2. is/was the difference in font-weight on _this_ specific component deliberate and/or important?

## Images

image

Left: existing implementation
Right: this PR's implementation

## Accessibility

My time on this provided by [Annertech](https://www.annertech.com/)

### DIFF
--- a/templates/_components/prev-next.twig
+++ b/templates/_components/prev-next.twig
@@ -7,6 +7,8 @@
 
 {% set previous_icon = previous_icon|default('chevron-left') %}
 {% set next_icon = next_icon|default('chevron-right') %}
+{% set previous_aria_label = 'Previous'|t ~ (previous_title ? ': ' ~ previous_title : '') %}
+{% set next_aria_label = 'Next'|t ~ (next_title ? ': ' ~ next_title : '') %}
 
 {%
   set classes = [
@@ -23,7 +25,7 @@
   <ul class="lgd-prev-next__list">
     {% if previous_url %}
       <li class="lgd-prev-next__list-item lgd-prev-next__list-item--prev">
-        <a href="{{ previous_url }}#lgd-guides__title" class="lgd-prev-next__link lgd-prev-next__link--prev" aria-label="{{ 'Previous'|t }}: {{ previous_title }}">
+        <a href="{{ previous_url }}#lgd-guides__title" class="lgd-prev-next__link lgd-prev-next__link--prev" aria-label="{{ previous_aria_label }}">
           {% include "@localgov_base/includes/icons/icon.html.twig" with {
             icon_name: previous_icon,
             icon_classes: 'lgd-prev-next__icon lgd-prev-next__icon--prev',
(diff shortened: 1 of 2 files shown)


msayoung:
Hi @ctorgalson - thanks for this. 

We do need the aria-labels to work, but I don't see the need for a specific template for it to do so, if we can get it working otherwise. Presumably we just need to set those variables in the template too?

` {% set previous_title = prev_step_title %} `

You probably know this, but the aria labels are there to give a better descriptive name and context for the link. 
https://www.w3.org/WAI/WCAG22/Understanding/link-purpose-in-context

ctorgalson:
@msayoung Ah, sorry, I think I wasn't specific enough in my comments. 

The _links_ retain their `aria-label` attributes,* but in this change the `<nav>` element loses _its_ `aria-label`.

As far as I can see, Step by Step is the only one of these Prev/Next `<nav>` elements (also used [in Guides](https://github.com/localgovdrupal/localgov_base/blob/1.x/templates/block/guides-prev-next-block.html.twig), and [apparently somewhere in Blogs](https://github.com/localgovdrupal/localgov_base/blob/1.x/templates/block/localgov-blogs-prev-next-block.html.twig)) that [has an `aria-label` attribute](https://github.com/localgovdrupal/localgov_base/blob/1.x/templates/views/views-view-list--localgov-step-by-step-navigation--prev-next.html.twig#L46).

Added later: the Publications `<nav>` (#564) has an `aria-labelledby` attribute that we have to consider as well.

But if we determine that the `<nav>` _ought_ to retain the `aria-label`, then it'll be straightforward to add it into [`templates/_components/prev-next.twig](https://github.com/localgovdrupal/localgov_base/blob/1.x/templates/_components/prev-next.twig) and provide suitable variables for all that template's various uses.

---
\* Looking at the generated output, I see they include a harmless but superfluous `:` in some circumstances. I've addressed that here too (e9c9ea8).

markconroy (approved):

markconroy:
This makes perfect sense to me. We have another place or two as well where we can start transitioning things over to this new prev/next pattern so it's easier to maintain.

---

Issue #312, opened 2022-10-25
Should we add a "Back to Top" button for long pages?


markconroy:
This might be a good feature to create for LGD and add it as a theme setting that can be switched on/off.

Perhaps we could add it only if you have scrolled past 1.5 times the height of the screen or something like that.

markconroy:
Completed, closing.

PR #595, opened 2024-08-26
312 add back to top button

ctorgalson:
## What does this change?

It adds a "Back to Top" link per #312 just after the end of `<main>`. I've set up variables in the component, and placed the twig template in `templates/_components`, anticipating that this might be a good candidate for SDC conversion.

From the Twig template's comments, the component works like this:

> The template includes two HTML elements:
> 
> 1. the "back-to-top" link itself
> 2. a "back-to-top-target" span
> 
> The javascript from localgov_base/back-to-top interacts with this template as follows:
> 
>  - absolutely positions the target span
>  - creates an intersection observer to observe it
>  - un-hides the link when the target enters the viewport
>  - re-hides the link when the target exits the *bottom* of the viewport
>  - re-hides the link when it's clicked [1]
> 
>  The template includes no wrapping <div> element, and the css explicitly sets the maximum size of the target to 1px x 1px to minimize the chances of an element from the template blocking content.
> 
> [1] on small enough screens--especially with several visible banners--the link will *already* be visible when #main-content is reached, so re-hiding the link fails in that circumstance.

## How to test

On any page that's at least 1.5x as long as the browser viewport ([the demo homepage works with a window > 1300px high):

- scroll normally, watching the bottom right corner of the viewport
- when the link appears, click it

- [ ] the link appears in the expected circumstances
- [ ] only the exact bounds of the link block other content
- [ ] clicking the link returns to `#main-content`
- [ ] the link re-hides on click (note: in circumstances where the link appears before/as soon as `#main-content` is visible, the link won't re-hide when clicked)
- [ ] the link re-hides when scrolled off the _bottom_ of the viewport subject to the same note as for click, above.

The main risks I can see are:

1. accidentally blocking content with the "target" span
2. confusion re: the new UI component

I think both of these are addressed by in CSS: (1) by explicitly restricting the size and location (1px x 1px, at extreme left of page), and (2) by use of an icon and LGD-matching link/button styles.

## Images

image

## Accessibility

The link uses `hidden="until-found"` for its hidden state. This means in supporting browsers (Chrome-based, at the moment), the link doesn't rely on javascript to be discovered (meaning keyboard navigation will reveal it even absent js).

### DIFF
--- a/css/components/back-to-top.css
+++ b/css/components/back-to-top.css
@@ -1,0 +1,6 @@
+.button.back-to-top:not([hidden]) {
+  bottom: 1em;
+  display: inline-block;
+  position: fixed;
+  right: 1em;
+}
--- a/css/components/back-to-top.css
+++ b/css/components/back-to-top.css
@@ -1,6 +1,35 @@
-.button.back-to-top:not([hidden]) {
-  bottom: 1em;
-  display: inline-block;
+.back-to-top {
+  background: var(--button-link-bg-color);
+  border: 3px solid var(--color-white);
+  bottom: 2em;
+  color: var(--button-link-color);
+  display: flex;
+  font-weight: bold;
+  gap: var(--spacing-small);
+  padding: var(--button-link-padding);
   position: fixed;
-  right: 1em;
+  right: 2em;
+  text-decoration: none;
+  transition: var(--transition-time);
+  opacity: 1;
+}
+
+.back-to-top[hidden] {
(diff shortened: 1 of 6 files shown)


markconroy (approved):
Looks good to me, thanks @ctorgalson.

I think I can see a few follow-up issues to make it a little bit more extensible perhaps, but definitely a great version 1 to add to our set up.

PR #598, opened 2024-08-29
592 move components to use new single directory components module

ctorgalson:
For the moment, this does the following:

1. declares `drupal:SDC` as a theme dependency
2. adds a replacement `localgov_base:prev-next` component at `components/prev-next/`
3. adds a replacement `localgov_base:icon` component at `components/icon/`
4. replaces theme template `icon.html.twig` with `localgov:icon` in new `localgov_base:prev-next` component _only_
5. replaces all `localgov_base` uses of `prev-next.html.twig` with the new `localgov_base:prev-next` component
6. marks "DEPRECATED" but **retains** `css/components/prev-next.css`
7. marks "DEPRECATED" but **retains** `templates/_components/prev-next.html.twig`

## How to test

Verify that no style or functional changes have crept-in. Mostly, this PR consists of moving twig templates and css, and updating other twig templates to reflect the new locations.

We can check that the expected markup/behaviour is present at content pages using each of the templates consuming the new component (obviously these changes don't exist on the live demo site):

- `templates/block/guides-prev-next-block.html.twig`:
  - [/order-copy-certificate/order-copy-certificate](https://demo.localgovdrupal.org/order-copy-certificate/order-copy-certificate)
- `templates/views/views-view-list--localgov-step-by-step-navigation--prev-next.html.twig`:
  - [/births-deaths-marriages-and-citizenship/marriage-and-civil-partnerships/step-by-step/get-married-4](https://demo.localgovdrupal.org/births-deaths-marriages-and-citizenship/marriage-and-civil-partnerships/step-by-step/get-married-4)
- `templates/navigation/book-navigation--publication.html.twig`:
  - [/publications/publications-cover-page-demo-content/petitions-scheme-publications-demo-content](https://demo.localgovdrupal.org/publications/publications-cover-page-demo-content/petitions-scheme-publications-demo-content)
- `templates/block/localgov-blogs-prev-next-block.html.twig`:
  - ???
  
The changes here _also_ anticipate that subthemes and other modules may be directly consuming the existing files and does _not_ remove them.

## Images

The markup and CSS are identical to what's currently in production.

## Accessibility

No changes of this kind have been made.


ctorgalson:
@markconroy here's my first attempt. The TL;DR is:

- I added the new component
- I **didn't** remove the existing components
- I created a replacement for `icon.html.twig` but only used it inside the new component (a future issue/PR could address the continuing use of `icon.html.twig` by `localgov_base`)
- we need to decide how to address the localgov disable css setting

markconroy:
Thanks Christopher. 

For the "Icon" component, is there any issue with sites (sub themes) that currently use the icon template? I'm guessing they are using something like

```
@include {% @localgov_base/inclues/icon.html.twig %}
```

But now that will update to something like 
```
@include {% localgov_base:icon with ... %}
```

Does that mean we are going to break current sites?

ctorgalson:
@markconroy That's why I left the original files in place with a big `DEPRECATION` comment at the top.

I could also update `icon.html.twig` to _use_ the new SDC.

ctorgalson:
@petrillek pointed out to me that I've been rashly passing variables through twig :roll_eyes:. I'll correct that on Monday.

ctorgalson:
This should stay in draft status for the moment: there's an issue with variable data types in `*.component.yml` that makes it easy to get a wsod.

ctorgalson:
> This should stay in draft status for the moment: there's an issue with variable data types in *.component.yml that makes it easy to get a wsod.

I've fixed that up, as well as converting the `add-to-calendar` template/library to a new SDC.

As with the other component, I've left the existing template and library files in place but added a `DEPRECATED` notice at the top of each file: so `localgov_base` doesn't currently use these, but sub-themes that do won't be impacted.

ctorgalson:
Not sure what's going on in that failing test; seems unrelated.

css/components/add-to-calendar.css
markconroy, line 1: Instead of marking the css/js files as deprecated, I wonder should we just update the library definitions for these and then remove the files altogether?

markconroy: We might be able to do it by something like this

```
add-to-calendar:
  css:
    theme:
      components/add-to-calendar/css/add-to-calendar.css
```

ctorgalson: That would be nicer. Don't see why it couldn't work. I'll try it out in a couple of minutes.

ctorgalson: Changed the libraries file to refer to the SDC files, and deleted the originals.

Tested by using `{{ attach_library('localgov_base/prev-next') }}` on a page with no prev-next navigation and verified the files were present.

stephen-cox:
Discussing at Merge Tuesday - with @markconroy to review

ctorgalson:
Thanks @markconroy. Been moving recently and lost the plot with these issues :)

markconroy:
No worries @ctorgalson 

It's a bit of a large issue to get the head back into after not looking at it for a while.

markconroy:
@ctorgalson I think this is ready to merge.

Do you want to give it a final test (especially as I have removed SDC as a dependency since it's now deprecated).
https://www.drupal.org/docs/core-modules-and-themes/deprecated-and-obsolete#s-single-directory-components-sdc

ctorgalson:
@markconroy The standalone module is deprecated, but don't we need to declare it as a dependency to ensure the core module is enabled?

markconroy:
I don't think so, it's now a service or something that in Drupal core, and not a module at all any more, so should "just work".  I think it's this file:
`web/core/lib/Drupal/Core/Render/Element/ComponentElement.php`

I removed it from our `localgov_base.info.yml` file and cleared the cache and all seemed fine.

markconroy (approved):
LGTM

---

finnlewis:
@markconroy we're a bit scared of this, it looks big! Waiting for your availability to discuss and liaise on the merge and release.

Sounds really cool though! Thanks @ctorgalson

markconroy:
It _is_ big. It could open up a whole new paradigm for our frontend system.

We are just testing it here with two components (icon and add-to-calendar) that we could easily revert if needed. Really, all that is happening is that the templates for those things are moving to somewhere else in the theme. 

In the future we can move more things, such as "Card view mode" so that if we wanted to have a card template for node and for user and for product and for paragraph etc, they could all extend from the same template. Pretty cool!

markconroy:
@ctorgalson Perhaps you could come along to Tech Group Drop-in on Wednesday to help talk through what we are doing with this PR?

ctorgalson:
@markconroy I can. Did it change from Thursday to Wednesday? (If so, can I get the Zoom link from you?)

markconroy:
You can register here, so it will pop into your calendar for you - https://lu.ma/g5nsdt8b

finnlewis:
To test, install demo on default install and check event content type: 

https://localgov.ddev.site/events/sports-player-year-awards-2022

finnlewis (approved):
Works for me locally!

PR #603, opened 2024-09-14
feat: add expand/collapse icon to accordion

### LINKED ISSUE #602: Add > or + to accordions by default
Our accordions look like buttons or calls to action. We should have an icon to denote that they expand/contract - a `+` or `>` or something like that.

https://demo.localgovdrupal.org/fostering/types-fostering

Screenshot 2024-09-13 at 09 14 50

millnut:
## What does this change?

Fixes #602


markconroy:
@millnut I've made some very small changes to your PR. Can you review it and approve if you are happy with it?

---

markconroy (approved):

millnut:
Yep looks all good to me @markconroy . Just a question for future is it best to use variables where possible for customisation flexibility or is there a general best practice for the frontend?

markconroy:
We are taking a "varibles-first" approach, so that if someone just wants some simple changes (e.g. to change the icons from `+` and `-` to `>` and `^`) they can just do so in their variables.css file.

For this specific issue, I'd have preferred to make these changes in the `html.twig` file, but given how accordions are set up mostly via JS, I'm happy to leave that for another day.

millnut:
@markconroy yes I guess there is a discussion on whether to change accordions to use native html `details and summary`

markconroy:
I don't think we have that discussion started yet, but I'd certainly welcome a chat about it.

PR #605, opened 2024-09-14
feat: add last updated date to service pages

millnut:
I've added this to some projects in the past, so I've rolled up this small change to a PR. This is probably one for further discussion as to whether this should be default. However if there is a case to not want this the template can be overridden.

## What does this change?

For service pages only this adds the date the page was last changed/updated

Example:
image

### DIFF
--- a/templates/content/node--full.html.twig
+++ b/templates/content/node--full.html.twig
@@ -146,6 +146,10 @@
       <div{{ content_attributes.addClass(content_type|clean_class ~ '__content', 'node__content') }}>
         {{ content|without('localgov_subsites_content') }}
       </div>
+
+      {% if content_type == 'localgov_services_page' %}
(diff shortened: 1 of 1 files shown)


markconroy:
I think this will definitely need some discussion, perhaps at Merge Tuesday or at a Tech Group Drop-in. 

It's adding a new feature to service pages only, but I have a feeling a number of councils may already have implemented something custom for this. We also have [an issue](https://github.com/localgovdrupal/localgov/issues/734) and [a PR](https://github.com/localgovdrupal/localgov/pull/735) to add this feature as a block (via views) to create a log of revisions.

---

markconroy:
@millnut Do you mind if we close this issue and we can open a more general issue if we need to so it's not tied to service pages. 

I think the issue in the localgov project might be more suited: https://github.com/localgovdrupal/localgov/issues/734

PR #617, opened 2024-09-23
ensures quick fact content is inside a <p> tag

### LINKED ISSUE #607: Missing semantic structure in quick facts section
### Description

The "Quick Fact" section on the service page lacks proper semantic structure. Key information is styled in a way that suggests importance but does not use appropriate semantic elements to convey its meaning.

### Expected Behaviour

- Information, structure, and relationships conveyed through presentation should be programmatically determined or available in text.
- Semantic elements like headings should be used to structure content properly.

### Actual Behaviour

- The "Quick Fact" section contains styled content that lacks semantic equivalents. For example, the main content is styled to appear larger but does not use appropriate semantic elements to indicate its significance.

### How to Replicate

1. Visit a service page, such as [this example page](https://a11y.localgovdrupal.org/adult-health-and-social-care/another-service-landing-page/service-page).
2. Locate the "Quick Facts" section.
3. Inspect the HTML markup to verify that the section uses div elements with styling but lacks semantic elements.

### WCAG Reference / Links

- **WCAG Guideline 1.3.1: Info and Relationships (A)**  
  *Information, structure, and relationships conveyed through presentation can be programmatically determined or are available in text, such as headings, landmarks, lists, tables, and footnotes.*

More details: [WCAG 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html)

### Recommendations

1. Update the "Quick Fact" section to use appropriate semantic HTML elements.

markconroy:
Closes #607 

## What does this change?

Our template "nearly" puts the text of a quick fact into `p` tags, but since we render `{{ content.localgov_fact }}` the render array produces a `div` which in turn means we render an empty `p` tag followed by the content.

This PR sorts that so that we render `{{ content.localgov_fact.value }}` instead so it goes inside the `p` tag.

## How to test

Visit https://localgov.ddev.site/adult-health-and-social-care/another-service-landing-page/service-page and check that the content in the fact box is inside a `p` tag.

## Accessibility
The original issue says "Key information is styled in a way that suggests importance but does not use appropriate semantic elements to convey its meaning.".

However, I am not sure what semantic elements to use except `p` for these elements. It's not clear _in all circumstances_ (so we could make this generic) if any of the constituent parts should be headings or something else.

---

### DIFF
--- a/templates/paragraphs/paragraph--localgov-fact-box.html.twig
+++ b/templates/paragraphs/paragraph--localgov-fact-box.html.twig
@@ -66,14 +66,14 @@
         {{ paragraph.localgov_above_text.value }}
       </p>
     {% endif %}
-    
+
     <div class="fact-box__fact">
       <p>
-        {{ content.localgov_fact }}
+        {{ paragraph.localgov_fact.value }}
       </p>
       {{ content|without('localgov_above_text', 'localgov_fact', 'localgov_below_text', 'localgov_background') }}
     </div>
-    
+
     {% if paragraph.localgov_below_text.value %}
       <p class="fact-box__closing">
         {{ paragraph.localgov_below_text.value }}


msayoung (approved):

msayoung:
Hi @markconroy , this is great thanks. 

I agree, its hard to know what to use which will make sense in all circumstances. I think generic  `<p>` elements are ok.

PR #637, opened 2024-10-15
Fix: Hide empty facets form controls

Adnan-cds:
Sorry Mark, this morning I forgot to test the scenario where no facets apply :(

## What does this change?
Empty facets forms should not display their control buttons.

## How to test
- Install the facets_form module.
- Edit the localgov_directories_facets facet from /admin/config/search/facets/localgov_directories_facets/edit
 - Select the `Checkboxes (inside form)` widget near the top of the page and save settings.
 - From the block layout page at /admin/structure/block, place the `Facet form: View Directory channel, display Embed block` in a sidebar.
 - [Optional] Disable the existing facet block from the block layout page at /admin/structure/block.
 - Open an existing directory channel page and search for an improbably string such as "foobarbazqux".
 - The "Search" and "Clear filter" control buttons of the Facets form will appear under the search form even though no facet items are applicable.

When searching for an improbable string, no part of the facets form should be visible.

Can't think of any.

## Images
image


Adnan-cds:
Replaced with #638

Issue #640, opened 2024-10-18
Prev/Next template adds a URL fragment that shouldn't be there.


rupertj:
Both the previous and next links rendered in the prev-next template add #lgd-guides__title to the URL.

Presumably this was fine for guides, but it's incorrect now that template's used for publications too.

PR #641, opened 2024-10-22
Removed URL fragment from template and moved it to a guides specific preprocess, so it only appears on guides.

rupertj:
Fix for #640


markconroy (commented):
LGTM

---

markconroy (approved):

rupertj:
Thanks! I probably won't be at MT, but if this could get in then, it'd be nice as it should stop the tests failing on publications :)

PR #645, opened 2024-11-07
adds housekeeping region

### LINKED ISSUE #644: Add housekeeping region
Most designs have region after all the footer blocks for housekeeping links - cookies, privacy, accessibility, etc.

Let's add a region for this so each council doesn't need to do it individually.

markconroy:
Closes #644 

## What does this change?

Adds a new region called "housekeeping", where people can put in links for things like cookies, privacy, etc.

## How to test

Checkout this branch, place a block in the housekeeping region, make sure you can see the block.

Fewer councils need to add their own region for this feature.

### DIFF
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -211,6 +211,13 @@ body {
   --color-post-footer-link: var(--color-link);
   --color-post-footer-text: var(--color-text);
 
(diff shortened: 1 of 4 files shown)


tanc (approved):
Looks good to me Mark. Thanks!

PR #646, opened 2024-11-07
Converts CSS to logical properties

### LINKED ISSUE #628: Use Locigal Properties for all our CSS
We could reduce a lot of our CSS - especially remove all our `[dir="rtl"]` type items if we used logical properties instead.

So ...

- instead of `padding-left`, we use `padding-inline-start`
- instead of `padding-right`, we use `padding-inline-end`
- instead of `margin-top`, we use `margin-block-start`
- etc

markconroy:
Closes #628 

## What does this change?

Removes all instances of things like `margin-left`, `margin-right`, etc and use logical properties for them instead, such as `margin-inline-start` and `margin-inline-end`. 

This will be very beneficial for any site that wants to use languages other than left-to-right languages. It will be even more beneficial if we scale the project beyond UK/Ireland. Oh, and it will also mean a smaller CSS codebase to maintain.

## How to test

- Checkout this branch
- Load the demo content module
- Make sure everything looks exactly as it did before

We only have to write one set of CSS rules, no matter what direction our language is in.

## Have you considered any issues?

This PR changes the names of all our properties such as `margin-left` but also all our custom variables such as `--quote-padding-left: var(--spacing-larger);` becoming `--quote-padding-inline-start: var(--spacing-larger);`.

### Solutions - 1: Release Notes
We will need to make sure we have release notes for people to update anywhere they have used those variables (it should be a simple enough find/replace task).

### Solutions - 2: Mapped variables
We could create new variables and map those to the old ones, such as:
- `--quote-padding-inline-start: var(--quote-padding-left);`

### DIFF
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -257,6 +257,7 @@ body {
 
   /* File Upload Items */
   --file-upload-border: var(--border);
+  --file-upload-border-left-size: var(--border-width-large); /* Deprecated */
   --file-upload-border-inline-start-size: var(--border-width-large);
   --file-upload-border-color: var(--color-accent);
   --file-upload-bg-hover-color: var(--color-accent);
@@ -321,6 +322,7 @@ body {
 
   /* Quote */
   --quote-padding: var(--spacing-larger);
+  --quote-padding-left: var(--spacing-larger); /* Deprecated */
   --quote-padding-inline-start: var(--spacing-larger);
   --quote-border: var(--border-large);
   --quote-border-color: var(--color-accent);
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -257,7 +257,7 @@ body {
 
   /* File Upload Items */
   --file-upload-border: var(--border);
-  --file-upload-border-left-size: var(--border-width-large);
+  --file-upload-border-inline-start-size: var(--border-width-large);
   --file-upload-border-color: var(--color-accent);
   --file-upload-bg-hover-color: var(--color-accent);
   --file-upload-text-hover-color: var(--color-white);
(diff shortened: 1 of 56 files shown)


css/components/page-sections.css
```diff
@@ -31,36 +31,36 @@
   --color-page-section-text-color: var(--color-page-section-background-color-6-text-color);
 }
 
-.lgd-page-section--padding-top-small {
-  padding-top: var(--spacing-small);
+.lgd-page-section--padding-block-start-small {
```
msayoung: Did you mean to update these class names, and if so are they updated in the templates too?

css/base/variables.css
```diff
@@ -257,7 +257,7 @@ body {
 
   /* File Upload Items */
   --file-upload-border: var(--border);
-  --file-upload-border-left-size: var(--border-width-large);
+  --file-upload-border-inline-start-size: var(--border-width-large);
```
msayoung, line 261: I'm not certain we want to change the variable names. Though I understand the reason for it... It would be good to discuss.

css/components/page-sections.css
markconroy: Oooh, good find. Let's not update the class names (and maybe in the next phase of this, we can add the new class names into the template as well as the current ones) and in a future phase, remove the current ones.

css/base/variables.css
markconroy: I think I'll go with this approach:

> Solutions - 2: Mapped variables

> We could create new variables and map those to the old ones, such as:

>    `--quote-padding-inline-start: var(--quote-padding-left);`

markconroy: @maria I think this is fixed now.

We now have
1. The old variable so we don't break existing sites, and
2. The new variable for how we will work from here on ...

msayoung (approved):

PR #660, opened 2024-11-24
fix: minor drupal-rector fixes

millnut:
## What does this change?

Just some minor drupal-rector fixes while I was running some lint/rector checks

### DIFF
--- a/tests/Functional/RegionRenderTest.php
+++ b/tests/Functional/RegionRenderTest.php
@@ -31,7 +31,7 @@ class RegionRenderTest extends BrowserTestBase {
   /**
    * Data provider for ::testRegionRender().
    */
-  public function blockRegionProvider() {
+  public static function blockRegionProvider() {
     return [
       [
         'region' => 'disabled',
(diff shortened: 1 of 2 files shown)


tests/Functional/RegionRenderTest.php
```diff
@@ -31,7 +31,7 @@ class RegionRenderTest extends BrowserTestBase {
   /**
    * Data provider for ::testRegionRender().
    */
-  public function blockRegionProvider() {
+  public static function blockRegionProvider() {
```
ekes, line 34: `: array`?

ekes (approved):
But might as well merge. I'm guessing there will be plenty more we should add return types and `strict` to.

millnut:
> But might as well merge. I'm guessing there will be plenty more we should add return types and `strict` to.

Yep these are just quick wins based on what drupal-rector changes, the rest are flagged as part of https://github.com/localgovdrupal/localgov_project/pull/179 in the baseline with the more thorough return type checking.

Issue #661, opened 2024-11-25
Banner primary: Link wraps all the banner contents


msayoung:
When a primary banner has a URL , the link `<a>` wraps the title, text and logo fields.
As all the child contents becomes part of the link name it becomes very verbose - and usually meaningless in terms of the link purpose.

I think we've taken this approach in order to make the whole area clickable, however there are better solutions for this.

An approach would be to make a visible element - such as the title - the actual link, and use some js to make the whole area clickable. However, as this has come up in the context of card design, we should probably converge on a solution. 

Notes. 
1. The URL field does have the following help text: "An optional url. Adding a url here will make the whole box a link." so if we change that behaviour we'll need to update this text. 

2. While I would generally recommend that the title of the Banner becomes the actual link text, it isn't obvious from the field layout that this will be the case. We'll need to make sure that this is clear for editors. 

3. We may not want to change the  behaviour on current sites without warning.

msayoung:
closing

PR #662, opened 2024-11-27
updates template/css so only the title becomes a link

markconroy:
Relates to #661 

## What does this change?

Changes the template for the primary banner paragraph type so that if there is a link for the banner, only the title becomes a link rather than all the content.

~I'll create a corresponding PR to update the help text for the field so it doesn't say the whole banner will be linked.~
[Corresponding PR for LocalGov Paragraphs](https://github.com/localgovdrupal/localgov_paragraphs/pull/205)

## How to test

- Create a subsite
- Add a Primary Banner to it
- Give that banner a link
- Give it a title
- Give it a summary
- Check that only the banner title is linked rather than all the content in the banner

This might make some changes to existing sites, so we need to communicate that in the release notes.

---

### DIFF
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -404,6 +404,8 @@ body {
   --banner-content-bg-color: var(--color-accent);
   --banner-content-text-color: var(--color-white);
   --banner-title-size: var(--font-size-larger);
+  --color-banner-title-link: var(--color-white);
+  --color-banner-title-link-hover: var(--color-white);
 
   /* Teaser View Mode */
   --teaser-image-with: 33%;
--- a/css/components/banner.css
+++ b/css/components/banner.css
@@ -22,11 +22,18 @@
   font-size: var(--banner-title-size);
 }
 
-.banner__link .field--name-localgov-title {
-  text-decoration: underline;
+.banner__link {
+  text-underline-offset: 5px;
+  text-decoration-thickness: 2px;
+  color: var(--color-banner-title-link);
 }
 
-.banner__link:focus .field--name-localgov-title,
-.banner__link:hover .field--name-localgov-title {
+.banner__link.banner__link:focus,
+.banner__link:hover {
   text-decoration: none;
+  color: var(--color-banner-title-link-hover);
+}
+.banner__link.banner__link:focus-visible {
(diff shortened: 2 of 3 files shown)


finnlewis:
I'm testing this now. Reading the original post #661 I'd be keen to add some magic (CSS / JS ?)to make the whole area clickable still.

finnlewis:
Could we do something along these lines? https://wpbeaches.com/fill-a-div-with-a-link-tag/ 

I guess it might be cleanest with some JavaScript to enhance, re-write the <a> tag, but would this then put us back in the original position of the (I assume) screen reader reading out the whole verbose contents of the link?

finnlewis:
@stephen-cox thinks he solved the anchor filling the box issue somewhere else recently...

markconroy:
We can add some simple CSS to make the whole thing clickable, quite easily.

```
a {
  position: relative;
}
a::after {
  position: absolute;
  inset: 0;
```

That should be enough CSS, and only the actual link text (i.e. the title) will be read out by screenreaders.

msayoung:
Why do we want the whole area to be clickable (in this situation)? 

I know this sort of thing (whole card clickabe) is wanted sometimes, I suggest we come up with our preferred methodology and then tell people how to apply it when they want it. Perhaps a spin out another issue? 

Note: the issue with Mark's suggestion is that it stops the text being selectable.

stephen-cox:
I solve this with JS recently using:

```js
const cards = document.querySelectorAll('.card');
Array.prototype.forEach.call(cards, card => {
  let down, up, link = card.querySelector('.card__title a');
  card.style.cursor = 'pointer';
  card.onmousedown = () => down = +new Date();
  card.onmouseup = () => {
    up = +new Date();
    if ((up - down) < 200) {
      link.click();
    }
  }
});
```

markconroy:
@stephen-cox That looks like pretty nice JS. Certainly worth exploring that a bit more.

@msayoung the original request here was to not have the whole banner text area wrapped in an `<a>` tag. Do you want to merge this PR so we have at least accomplished that? Then we can figure out our approach to "whole-card-clickable" later?

msayoung (approved):

PR #694, opened 2024-12-18
Adds toggle to use stacked header for guides

### LINKED ISSUE #675: add 'use stacked header on guides' as an localgov_base option, default to off
Follow up from https://github.com/localgovdrupal/localgov_guides/issues/122#issuecomment-2528304495

markconroy:
Closes #675 

## What does this change?

Adds a toggle on the theme settings page to allow people to set "Use stacked header" for guides.

If this is set, then we can use the `{{ subtitle }}` in the page header Twig file to use the stacked header pattern.

The only problem is, guides does not have a field for subtitle! So this is WIP until that happens.

---


markconroy:
This is probably superceded by #716 

If #716 gets merged, we can then close this one.

markconroy:
#716 is approved, let's close this.

PR #704, opened 2025-01-23
fix: stylelint errors

millnut:
## What does this change?

Fixes stylelint errors

### DIFF
--- a/components/prev-next/prev-next.css
+++ b/components/prev-next/prev-next.css
@@ -15,8 +15,7 @@ ul.lgd-prev-next__list li.lgd-prev-next__list-item {
   display: flex;
   flex-wrap: wrap;
   align-items: center;
-  padding: var(--button-padding-vertical) var(--button-padding-horizontal)
-    var(--button-padding-vertical) var(--button-padding-horizontal);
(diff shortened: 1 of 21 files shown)


markconroy:
@millnut This PR looks really strange. I prefer the way the CSS is in this PR - items remaining on one line - but they way it currently is before this PR was also created via stylelint. 

Has the stylelint rules changed to put these back to one line?

millnut:
Hey @markconroy I'm not sure it uses this rule; 

https://git.drupalcode.org/project/drupal/-/blob/10.4.x/core/.stylelintrc.json?ref_type=heads

Looking at the last [failed test](https://github.com/localgovdrupal/localgov_base/actions/runs/12931898124/job/36067012893) stylelint is auto-detecting the rules for prettier and including those as well, which Core then specifcally declare long lines for CSS https://git.drupalcode.org/project/drupal/-/blob/10.4.x/core/.prettierrc.json?ref_type=heads `"printWidth": 10000,` vs other files types `"printWidth": 80,`.

Not really sure on the reason for the Core decisions of `10000` characters for only CSS files

markconroy:
Well, if the documented rule states that and the linter followed the rule, let's get this approved.

markconroy (approved):

PR #706, opened 2025-01-23
Fix/1.x/705 lgd icon doesnt use theme variable

### LINKED ISSUE #705: lgd:icon SDC doesn't use theme variable
The `lgd:icon` SDC [accepts a `theme` variable](https://github.com/localgovdrupal/localgov_base/blob/1.x/components/icon/icon.twig#L35), but [doesn't make any use of it](https://github.com/localgovdrupal/localgov_base/blob/1.x/components/icon/icon.twig#L45). This makes it impossible to use any icon from outside of `localgov_base`, largely defeating the purpose of the SDC :)

~This is somewhat mitigated by the fact that the variable was only added when I converted the existing Twig template to an SDC,~ and [is not documented](https://github.com/localgovdrupal/localgov_base/blob/1.x/components/icon/icon.component.yml).

PR to follow shortly.

ctorgalson:
Closes #705

## What does this change?

The PR adds the following to `localgov_base:icon`:

1. use the `theme` variable to namespace the path to the icon file if `icon_path` does not begin with `@`,
2. document that the `theme` variable exists in `icon.component.yml` (it was already documented in code),
3. adds a `|default('includes/icons')` to the `icon_path` variable to make the SDC behave as described in comments

## How to test

1. create a child theme, `lgdtest`,
2. add an icon like `web/themes/custom/lgdtest/assets/icons/someIcon.svg` (preferably different from any icon in `localgov_base`),
4. include the `localgov_base:icon` component in the new child theme in _each_ of the following ways: 
    ```twig
    {# Test icon from another theme, using 'theme' var #}
    {% include 'localgov_base:icon' with {
      theme: 'lgdtest',
      icon_path: 'assets/icons',
      icon_name: 'someIcon',
    } %}
    
    {# Test icon from another theme, namespacing 'icon_path' var (note: `theme` var will be ignored) #}
    {% include 'localgov_base:icon' with {
      theme: 'lgdtest',
      icon_path: '@lgdtest/assets/icons',
      icon_name: 'someIcon',
    } %}
    
    {# Test icon from localgov_base, using 'theme' var #}
    {% include 'localgov_base:icon' with {
      theme: 'localgov_base',
      icon_name: 'arrow-down',
    } %}
    
    {# Test icon from localgov_base, namespacing 'icon_path' var (note: `theme` var will be ignored) #}
    {% include 'localgov_base:icon' with {
      theme: 'localgov_base',
      icon_path: '@localgov_base/includes/icons',
      icon_name: 'arrow-down',
    } %}
    
    {# Test icon from localgov_base, using *only* icon_name var #}
    {% include 'localgov_base:icon' with {
      icon_name: 'arrow-down',
    } %}
    ```
5. verify that the expected icons _all_ appear.

- check that _existing_ consumers of `localgov_base:icon` (i.e. [`localgov_base:prev-next`](https://github.com/localgovdrupal/localgov_base/tree/1.x/components/prev-next))work as expected
- verify all the anticipated modes (above) render correctly
- this _should_ leave existing implementations untouched

- the potential risks are wsod on templates making use of the `localgov_base:icon` SDC, so testing matters :)

## Images

n/a

## Accessibility

n/a

### DIFF
--- a/components/icon/icon.twig
+++ b/components/icon/icon.twig
@@ -35,6 +35,10 @@
 {% set theme = theme|default('localgov_base') %}
 {% set icon_wrapper_element = icon_wrapper_element|default('div') %}
 
+{%- set icon_fullpath -%}
+  {% if icon_path|first != '@' %}@{{ theme }}/{% endif %}{{ icon_path }}/{{ icon_name }}.svg
+{%- endset -%}
+
 {% set attributes = create_attribute()
   .setAttribute('aria-hidden', decoration|default('true'))
   .addClass('lgd-icon')
@@ -42,5 +46,5 @@
 %}
 
 <{{ icon_wrapper_element }}{{ attributes }}>
-  {% include icon_path ~ '/' ~ icon_name ~ '.svg' %}
+  {% include icon_fullpath %}
 </{{ icon_wrapper_element }}>
--- a/components/icon/icon.twig
+++ b/components/icon/icon.twig
@@ -36,7 +36,7 @@
 {% set icon_wrapper_element = icon_wrapper_element|default('div') %}
 
 {%- set icon_fullpath -%}
-  {% if icon_path|first != '@' %}@{{ theme }}/{% endif %}{{ icon_path }}/{{ icon_name }}.svg
+  {% if icon_path|first != '@' %}@{{ theme }}/{% endif %}{{ icon_path|default('includes/icons') }}/{{ icon_name }}.svg
 {%- endset -%}
 
 {% set attributes = create_attribute()
--- a/components/icon/icon.twig
+++ b/components/icon/icon.twig
@@ -35,10 +35,6 @@
 {% set theme = theme|default('localgov_base') %}
 {% set icon_wrapper_element = icon_wrapper_element|default('div') %}
 
-{%- set icon_fullpath -%}
-  {% if icon_path|first != '@' %}@{{ theme }}/{% endif %}{{ icon_path|default('includes/icons') }}/{{ icon_name }}.svg
-{%- endset -%}
-
 {% set attributes = create_attribute()
   .setAttribute('aria-hidden', decoration|default('true'))
   .addClass('lgd-icon')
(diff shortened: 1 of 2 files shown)


components/icon/icon.twig
```diff
@@ -35,12 +35,16 @@
 {% set theme = theme|default('localgov_base') %}
 {% set icon_wrapper_element = icon_wrapper_element|default('div') %}
 
+{%- set icon_fullpath -%}
+  {% if icon_path|first != '@' %}@{{ theme }}/{% endif %}{{ icon_path|default('includes/icons') }}/{{ icon_name }}.svg
```
markconroy: I think we can remove the `{% set ... %}` here then replace line 49 later with line 39 from here.

components/icon/icon.twig
```diff
@@ -35,12 +35,16 @@
...
+
 {% set attributes = create_attribute()
   .setAttribute('aria-hidden', decoration|default('true'))
   .addClass('lgd-icon')
   .addClass(icon_classes)
 %}
 
 <{{ icon_wrapper_element }}{{ attributes }}>
-  {% include icon_path ~ '/' ~ icon_name ~ '.svg' %}
+  {% include icon_fullpath %}
```
markconroy: Let's change this line to 
`{% if icon_path|first != '@' %}@{{ theme }}/{% endif %}{{ icon_path|default('includes/icons') }}/{{ icon_name }}.svg`

_after_ we remove the `set` from above, like so:

```
<{{ icon_wrapper_element }}{{ attributes }}>
  {% if icon_path|first != '@' %}@{{ theme }}/{% endif %}{{ icon_path|default('includes/icons') }}/{{ icon_name }}.svg
</{{ icon_wrapper_element }}>
```

markconroy (changes requested):
Looks good to me. I've added a small suggestion that we can accept or reject, it won't make much of a difference either way.

components/icon/icon.twig
ctorgalson: That will be nicer. I'll do that this afternoon.

finnlewis:
@ctorgalson I think if you merge 1.x back to your forked branch, the eslint checks will pass.

components/icon/icon.twig
ctorgalson: "This afternoon" came today for me :roll_eyes:

markconroy (approved):

PR #729, opened 2025-03-18
allowing ID randomization but keeping the stable selector

PardeepSinghBali:
## What does this change?

This PR addresses an issue related to form container ID randomization for improved accessibility and Address Dropdown not populating the addresses at drupal Form. Previously, the localgov_base_preprocess_container() function was appending a random suffix to the edit-actions form container ID. This introduced potential issues with JavaScript selectors that rely on the #edit-actions ID, as they were no longer able to target the element due to the dynamic nature of the ID.

**This change ensures that:**

1. The ID is randomized for accessibility, but we still maintain a stable selector (data-drupal-selector) to be used by JavaScript.
2. JavaScript selectors that depend on the original ID (edit-actions) will work as expected, avoiding potential breakage in behavior.
3. The random ID is applied to the id attribute, but the original ID is retained as data-drupal-selector for stable targeting.

## How to test

Check the edit-actions form container element:

On the main branch, inspect the container and check the id attribute. It should be static (e.g., #edit-actions).
On this branch, inspect the same container and verify that the id has been randomized (e.g., #edit-actions--<random>) but the data-drupal-selector remains as edit-actions.

## How can we measure success?

1. No errors or broken JavaScript functionality related to the #edit-actions container ID after the change.
2. JavaScript selectors that use data-drupal-selector="edit-actions" work as expected. 
3. Improved accessibility with randomized form container IDs while keeping the stable selector intact.

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

Issue #324, opened 2022-12-01
Newsroom search submit button has same id as sitewide search


msayoung:
https://site1.demo.microsites.localgovdrupal.org/newsroom 

```
<div data-drupal-selector="edit-actions" class="form-actions js-form-wrapper form-wrapper" id="**edit-actions**"><input data-drupal-selector="edit-submit-localgov-news-search" type="submit" id="edit-submit-localgov-news-search" value="Apply" class="button js-form-submit form-submit">
</div>

<div data-drupal-selector="edit-actions" class="form-actions js-form-wrapper form-wrapper" id="**edit-actions**"><input data-drupal-selector="edit-submit-localgov-sitewide-search" type="submit" id="edit-submit-localgov-sitewide-search" value="Apply" class="button js-form-submit form-submit">
</div>
```

msayoung:
Also a problem here: https://demo.localgovdrupal.org/news

msayoung:
https://github.com/localgovdrupal/localgov_base/pull/325

PR #731, opened 2025-03-19
Fixing ID randomization but keeping the stable selector

PardeepSinghBali:
## What does this change?

This PR addresses an issue related to form container ID randomization for improved accessibility and Address Dropdown not populating the addresses at drupal Form. Previously, the localgov_base_preprocess_container() function was appending a random suffix to the edit-actions form container ID. This introduced potential issues with JavaScript selectors that rely on the #edit-actions ID, as they were no longer able to target the element due to the dynamic nature of the ID.

This change ensures that:

The ID is randomized for accessibility, but we still maintain a stable selector (data-drupal-selector) to be used by JavaScript.
JavaScript selectors that depend on the original ID (edit-actions) will work as expected, avoiding potential breakage in behavior.
The random ID is applied to the id attribute, but the original ID is retained as data-drupal-selector for stable targeting.

## How to test

Check the edit-actions form container element:

On the main branch, inspect the container and check the id attribute. It should be static (e.g., #edit-actions).
On this branch, inspect the same container and verify that the id has been randomized (e.g., #edit-actions--) but the data-drupal-selector remains as edit-actions.

No errors or broken JavaScript functionality related to the #edit-actions container ID after the change.
JavaScript selectors that use data-drupal-selector="edit-actions" work as expected.
Improved accessibility with randomized form container IDs while keeping the stable selector intact.
## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -301,7 +301,7 @@ function localgov_base_views_pre_render(ViewExecutable $view): void {
 /**
  * Implements hook_preprocess_container().
  */
- function localgov_base_preprocess_container(&$variables): void {
+function localgov_base_preprocess_container(&$variables): void {
   // See https://www.drupal.org/project/drupal/issues/1852090
   // Ensure the container has an ID.
   if (isset($variables['element']['#id'])) {
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -301,19 +301,23 @@ function localgov_base_views_pre_render(ViewExecutable $view): void {
 /**
  * Implements hook_preprocess_container().
  */
-function localgov_base_preprocess_container(&$variables): void {
-
-  // Randomize form container IDs for accessibility.
+ function localgov_base_preprocess_container(&$variables): void {
   // See https://www.drupal.org/project/drupal/issues/1852090
+  // Ensure the container has an ID.
   if (isset($variables['element']['#id'])) {
-    $id = $variables['element']['#id'];
-    if ($id === 'edit-actions') {
-      $id .= '--' . Crypt::randomBytesBase64(8);
+    $original_id = $variables['element']['#id'];
+
+    // Only modify 'edit-actions' to prevent affecting other elements.
+    if ($original_id === 'edit-actions') {
+      $random_suffix = Crypt::randomBytesBase64(8);
+      $random_id = $original_id . '--' . $random_suffix;
+
+      // Keep a stable selector but allow ID randomization.
+      $variables['attributes']['id'] = $random_id;
+      $variables['attributes']['data-drupal-selector'] = $original_id;
+      $variables['element']['#attributes']['data-drupal-selector'] = $original_id;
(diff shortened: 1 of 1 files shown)


PardeepSinghBali:
Discussion can be followed from previous Closed PR: https://github.com/localgovdrupal/localgov_base/pull/729

andybroomfield:
Have tested and this works with the address lookup webform element still.

Have noticed that with this fix the edit-actions doesn't get the collision aviodance random ID if it has a --3 etc already. But I think this can still cause a problem so I would suggest applying it too all containers that start with `edit-actions`. 
See the original issue #324 and PR #325.

Screenshot 2025-03-19 at 2 41 46 pm

Screenshot 2025-03-19 at 2 52 34 pm

markconroy:
I think this looks good to me. I don't think we are using ids anywhere in our JS/CSS for styling (which in any case we couldn't since they would be changing each time we reload a page).

@andybroomfield you've been giving this a good test. If you are okay with how it's working, then I'm happy for you t approve it.

andybroomfield:
@markconroy I think its good, just the fix to handle use cases where the id starts with `edit-actions` as the original issue is the ID collision could still happen in those cases when the block is cached. /cc @PardeepSinghBali

finnlewis:
Discussed briefly at Merge Tuesday, sounds like it needs a little more work.

@PardeepSinghBali are you OK to review ?

PardeepSinghBali:
> Discussed briefly at Merge Tuesday, sounds like it needs a little more work.
> 
> @PardeepSinghBali are you OK to review ?

Yes sure, I will

PardeepSinghBali:
> Have tested and this works with the address lookup webform element still.
> 
> Have noticed that with this fix the edit-actions doesn't get the collision aviodance random ID if it has a --3 etc already. But I think this can still cause a problem so I would suggest applying it too all containers that start with `edit-actions`. See the original issue #324 and PR #325.
> 
> Screenshot 2025-03-19 at 2 41 46 pm
> 
> Screenshot 2025-03-19 at 2 52 34 pm

done and changes pushed

andybroomfield (approved):
Thanks @PardeepSinghBali this is working as expected, and the id randomisation is applied to all iterations of edit-actions.

PardeepSinghBali:
> Thanks @PardeepSinghBali this is working as expected, and the id randomisation is applied to all iterations of edit-actions.

Perfect!

PR #734, opened 2025-03-20
Move restricted_width_content_types to .theme file

### LINKED ISSUE #733: Move restricted_width_content_types to localgov_base.theme
It's hard (impossible?) to add to the array of `restricted_width_content_types` without overriding the `node--full.html.twig` template. This is a big price to pay if you just want to add another content type to the list, as it means any updates to `node--full-html.twig` will not longer be received by your subtheme.

If we move that array to the `localgov_base.theme` file, then we can add to the array from our `subtheme.theme` file and leave the `node--full.html.twig` file in place.

markconroy:
Closes #733 

## What does this change?
Moves the declaration of the `restricted_widths_content_types` from `node--full.html.twig` to `localgov_base.theme` so it's easier for subthemes to add to/remove from the array.

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -197,7 +197,7 @@ function localgov_base_preprocess_page(&$variables): void {
  * @return array
  *   An array of restricted width content types.
  */
-function _get_restricted_width_content_types() {
+function _localgov_get_restricted_width_content_types() {
   return [
     'localgov_services_page',
     'localgov_event',
@@ -219,7 +219,7 @@ function localgov_base_preprocess_node(&$variables): void {
     // some content types. This is helpful to reduce the line-length for
     // better readability.
     $variables['restricted_width_content_types'] = [];
-    if (in_array($node->bundle(), _get_restricted_width_content_types())) {
+    if (in_array($node->bundle(), _localgov_get_restricted_width_content_types())) {
       $variables['restricted_width_content_types'][] = $node->bundle();
     }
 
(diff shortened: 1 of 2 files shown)


finnlewis:
@tonypaulbarker would like to test this.

markconroy:
@tonypaulbarker Any update on testing this? I'd like to get it accepted/rejected as soon as possible so I can keep working on the full view mode template for SDC.

tonypaulbarker:
@markconroy half time report is this: I like the idea, I think it'll work, yet to carry out functional testing. 

My early concern was for someone who has overridden the template in their sub-theme. Having examined the code I think it's going to be okay. If the array has been set in the template with a copy of the old code then those should still be the values used. And if the class isn't being used at all that's not a case we need be concerned about either.

markconroy:
@tonypaulbarker Great stuff, thanks. Can you mark this as approved when you get a chance to finish your testing? Thanks.

tonypaulbarker:
@markconroy it doesn't seem as though I have the permissions to approve this but approved subject to a check / remedy if needed of the two failing automated tests.

It seems events have been missing this class as they are using a specified template that doesn't set the variable. With the new code events do get the class. Not a blocker but worth noting for release notes.
I found some instances of content widths too narrow or too wide around the demo site already present and not related to this change. I'll file a new issue to look at those.

markconroy:
Thanks @tonypaulbarker, that's merged now.

Issue #600, opened 2024-09-03
LGD components: consolidate?


ctorgalson:
Disclaimer: I don't know if this issue belongs with `localgov_base`, but since it's the main and default frontend for the project I'm starting this here.

Having worked on theming a number of LGD projects, I find there's a bit of friction when customizing essentially identical components from different modules (and the theme itself) that, while well-suited to their original use, aren't compatible with _other_ each other.

Mainly, I'm thinking about grid systems--I believe I counted _four_ grid systems in use from LGD sources on one project--and widgets that implement [WAI-ARIA patterns](https://www.w3.org/WAI/ARIA/apg/patterns/) such as `localgov_step_by_step`'s and `localgov_paragraphs`' accordions.

So:

- does this cause enough friction for others that we should think about tackling it when we find obvious duplication of function?
- if so how (and where) should we start to go about it? (I.e. would we override modules' implementations in the theme, or provide SDCs or what-have-you in a module for consumption by themes _and_ modules?)

andybroomfield:
I think a lgd components module (maybe in localgov_core) could be a useful way of introducing single directory components if we want to go down that route. What is the Drupal core version where they became avalible?

ctorgalson:
@andybroomfield it's been in core as 'experimental' since Drupal 10.1, and 'stable' since 10.3.

I'm partial to the idea of a module as well, since that would make it available to LGD (and other) modules and `localgov_base` (and other themes) as an optional requirement if needed/wanted.

markconroy:
@ctorgalson The reason there is "stuff" in modules and other "stuff" in the theme is because the modules are supposed to be able to be standalone. To make this happen, we were providing templates + css + js in modules, and then overriding all that in the base theme.

We are now at a stage where we only provide basic markup/CSS in (new) modules and let LocalGov Base do most of the heavy lifting.

I think it's a really good idea for us to consolidate what we can - in the base theme or in a module. I kind of prefer doing it at the theme-level, but that might be just because I know the theme so well. Let's not let that stop us.

ctorgalson:
> The reason there is "stuff" in modules and other "stuff" in the theme is because the modules are supposed to be able to be standalone. To make this happen, we were providing templates + css + js in modules, and then overriding all that in the base theme.

Yep. As I mentioned, well-suited to doing the jobs they were built for.

[...]

PR #736, opened 2025-03-20
Update node--full to use SDC

### LINKED ISSUE #735: Convert node--full.html.twig to use Single Directory Components
Parent: #732

markconroy:
Closes #735 

Do not merge this until we merge #734 - if we do not merge #734 then we need to add the restricted_width-content_types array into the components/full.twig file.


finnlewis:
Discussing in Merge Tuesday, @tonypaulbarker suggests there is more work needed here before merging.

Let's review!

Can we add some visual regression tests to check this sort of thing doesn't have detrimental effects/affects?

markconroy:
@finnlewis sorry I should have marked this as WIP. I still have to fill out the schema for the full.component.yml file and a few other bits.

tonypaulbarker:
@markconroy some suggestions, keeping in mind the SDC ecosystem like https://www.drupal.org/project/sdc_display and storybook goals - some of it is thinking out loud, I've yet to try component-ising full node content.

Consider naming such as 'display-full' or 'view-mode-full' or 'content-full' rather than 'full'. People working in the UI in future  or including the component in their twig will not have the context of the directory.

I think props and slots should be defined in the component yml file. Particularly where we are drilling down into an array, e.g. node.isPromoted() I think should become something like full_content_is_promoted or is_promoted. Looking at other components we have so far, atc_start_date has not been specified but is used in https://github.com/localgovdrupal/localgov_base/blob/1.x/components/add-to-calendar/add-to-calendar.twig . I personally like to declare them all so we can do things like mark fields as required, map them in future in Drupal configuration and see a reference. 

The component should possibly be embedded rather than included, with slots like full_content embedded using blocks so that it'd be easy to mark up some demo content in storybook. I think it'll work in its current form, though.

The full.css file is empty. I think the CSS should be added there?

The SDC's CSS should be auto included by SDCs so we should check the localgov_base_remove_css does what is expected although in practice there are some issues with ordering ( https://www.drupal.org/project/drupal/issues/3374901 and https://www.drupal.org/project/drupal/issues/3046636 ) that may complicate what we observe. 

I think the step-by-step and events libraries CSS files should move to this component. Maybe there is a way that we can declare each of the libraries in the component.yml. 

Hope this helps!

markconroy:
Howdy @tonypaulbarker, thanks a lot for your thoughts in this thread. 

My plan for moving things to SDC (and making the path towards a design system easier) is to do it in steps. I'm not fully sure of the steps but something like this:

1. Move the templates only. This will mean we are only interested in the HTML being a component and our node/paragraph/etc templates being able to map to them with no change in the outputted HTML. This is the MVP for me for moving to SDC.
2. Move the assets - CSS + JS. This is a little bit trickier if we need to consider the `{% if not remove_css %} {{ add_library }} {% endif %}`. Since SDC automatically creates a library for us, I'm not sure how to tell our templates how to exclude that library if our `remove_css` option is checked. I'm happy to figure this out as part of a follow-up for each component.
3. Create the individual schemas my-component.component.yml files. This is a required file, but can be empty. I'm happy to leave these as empty for now and follow-up with a task to create these later.

To your specific comments:

> Consider naming such as 'display-full' or 'view-mode-full' or 'content-full' rather than 'full'. People working in the UI in future or including the component in their twig will not have the context of the directory.

Good idea, done.

> I think props and slots should be defined in the component yml file. Particularly where we are drilling down into an array, e.g. node.isPromoted() I think should become something like full_content_is_promoted or is_promoted. Looking at other components we have so far, atc_start_date has not been specified but is used in https://github.com/localgovdrupal/localgov_base/blob/1.x/components/add-to-calendar/add-to-calendar.twig . I personally like to declare them all so we can do things like mark fields as required, map them in future in Drupal configuration and see a reference.

I'm happy to leave this to phase 3 (if you/someone else wants to get it done without holding up phase 1, I'm fine with that)

> The component should possibly be embedded rather than included, with slots like full_content embedded using blocks so that it'd be easy to mark up some demo content in storybook. I think it'll work in its current form, though.

The variables we have in this specific component are direct 1:1 items. I don't think we need to embed here since we are not using any `{% block %}` items. We can swap to embed if we need to later/need to in other components.

> The full.css file is empty. I think the CSS should be added there?

Check out my note above about leaving this to phase 2 of this project.

> The SDC's CSS should be auto included by SDCs so we should check the localgov_base_remove_css does what is expected although in practice there are some issues with ordering ( https://www.drupal.org/project/drupal/issues/3374901 and https://www.drupal.org/project/drupal/issues/3046636 ) that may complicate what we observe.

Currently, if we have the CSS in the component and we turn off the CSS from the theme via the theme settings, the CSS for all the SDC components is still included. Let's figure this out as a phase 2 item.

> I think the step-by-step and events libraries CSS files should move to this component. Maybe there is a way that we can declare each of the libraries in the component.yml.

We'll figure out the CSS as part of phase 2. Events will not need to move here as events has it's own template. I've removed the bit that adds the events library, it was added from before events had it's own template which calls in that library.

finnlewis:
Discussing in Merge Tuesday, @markconroy thinks this is ready for testing again.

@andybroomfield asked about backwards compatibility. Confirming that if people have existing template overrides, these will not be affected.

tonypaulbarker:
@finnlewis @andybroomfield we can test it but theory says it should work fine as long as Single Directory Components has been enabled. 

@markconroy I don't see the component.yml causing much of a problem ahead but I am keen to understand our possible strategies for handling the remove_css parameter or we might be stuck at templates moved to no real end. How else could the switch be handled? Can some custom module detect the switch and help us ignore the assets in SDCs? Do we know which sites make use of it? Could a separate localgov_base_no_css theme be maintained alongside this one?

markconroy:
> as long as Single Directory Components has been enabled.

It's not a module any more. It's just part of the frontend theming system so cannot be turned on or off. It's _just there always_.

> I don't see the component.yml causing much of a problem ahead but I am keen to understand our possible strategies for handling the remove_css parameter or we might be stuck at templates moved to no real end. How else could the switch be handled? Can some custom module detect the switch and help us ignore the assets in SDCs? Do we know which sites make use of it? Could a separate localgov_base_no_css theme be maintained alongside this one?

@tonypaulbarker Is that question a blocker to approving this PR. If not, let's get this approved and we can figure that out next.

tonypaulbarker:
@markconroy you're absolutely right - that module is just a stub now.

Let's not be blocked and wait to figure out assets. Factoring out the component.yml and twig is going to be a big chunk of the work and we can be getting that work done.

tonypaulbarker:
@markconroy unfortunately in testing I did find a blocker. During the transition there _is_ the potential for breaking people's stuff. If someone copies the Drupal template but not the SDC then when the SDC and template are later updated in localgov_base the props, slots and blocks won't match and there will be issues.

markconroy:
@tonypaulbarker I don't see that as a blocker. If someone copies the template (in the future) they are in the same position as if they copy the template now. Once you copy the template you have decided to forfeit the updates that the template gets - that's a trade off every website in Drupal makes - custom template with full control vs not so custom but free updates.

If someone copies the (new version) of our system, it's up to them to update/maintain their SDC if they wish.

If that's the only blocker, I think we can mark this as approved?

tonypaulbarker:
@markconroy the difference here is that there will inevitably be breaking changes in the case where the template has been copied whereas up to this point someone could expect that once the template had been copied the template will go on working (passing up on updates). We _could_ communicate the change with comments but I think we should make a decision about where the components will ultimately end up so we don't find we reverse it later, see https://github.com/localgovdrupal/localgov_base/issues/600

markconroy:
I'm not sure anyone will copy the template on its own any more. There would be no point in doing that, since the template just maps Drupal variables to SDC variables. It's the SDC component you would need to copy/change (and perhaps the template _then_ if you were going to add new variables into the SDC for your custom implementation). But that's the same basically as what we currently do - copy the template, then it's up to you to maintain it.

Could you come along to Merge Tuesday tomorrow to explain it? Maybe I'm missing something.

The issue #600 is a slightly different issue.

tonypaulbarker:
@markconroy sure I'll be on the call. 

You're right that you can't really make effective changes this way but I expect this user to be passively copying files over to their sub theme and to have never met SDCs before. I think we can help avoid confusion and breaking changes for people by adding a bit of inline documentation in the comments.

markconroy:
> I think we can help avoid confusion and breaking changes for people by adding a bit of inline documentation in the comments.

Yes, exactly. That's why I don't see this as a blocking issue, it's just an issue of documentation.

finnlewis:
Discussing in Merge Tuesday...

There is related discussion in recent days 
 - https://github.com/localgovdrupal/localgov_core/issues/280
 - https://github.com/localgovdrupal/localgov_core/pull/281

@tonypaulbarker is happy to proceed in this direction as long as we realise that we can only get so far. For example, we won't be able to move the css and js into the SDC. One concern is that if people have overridden libraries, these overrides might change if we move libraries to SDC.

@markconroy is keen to merge this to bring the markup / templates across without necessarily moving js and css (yet?)

@tonypaulbarker is happy with this!

Also see https://github.com/localgovdrupal/localgov_base/issues/600#issuecomment-2783657214 for wider discussion on direction.

finnlewis (approved):

PR #740, opened 2025-03-26
Extend preprocess block/node to recognise both draft and archived states

anthonylindsay:
## What does this change?

Currently, you can optionally turn on a thing which will prepend `[draft]` to a node title, both in preprocess_node and preprocess_block. This change recognises moderation states so it will also label an archived node as archived.

## How to test

1. turn on the 'draft label' setting in theme settings
2. Unpublish a node. See that it says '[draft]' before the node title
3. Publish the node. See that the prepended label is gone.
4. Archive the node. See that it now says '[Archived]' before the node title

Not sure about that.

It assumes moderation states exist, so I guess the risk would be if someone turns off Workflow then it would break. The workaround would be to turn off the feature in theme settings.

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -201,9 +201,15 @@ function localgov_base_preprocess_node(&$variables): void {
       if ($state == 'draft') {
         $variables['label'] = t('[Draft]') . " " . $variables['node']->label();
       }
-      elseif ($state == 'archived') {
-        $variables['label'] = t('[Archived]') . " " . $variables['node']->label();
-      }
+    }
+  }
+  // We have a theme setting to prepend 'archived' to the title of
+  // archived content, but we don't mind whether it is published or not.
+  // An archived state does not have to always be unpubilshed.
+  if (theme_get_setting('localgov_base_add_archived_note_to_archived_content') === TRUE) {
+    $state = $variables['node']->get('moderation_state')->getString();
+    if ($state == 'archived') {
+      $variables['label'] = t('[Archived]') . " " . $variables['node']->label();
     }
   }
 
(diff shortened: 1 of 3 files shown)


markconroy (changes requested):
The "add Draft" is created via a theme setting in LocalGov Base. 

I think we should do the same for this "Archived" setting.

Some councils may want to use "Draft" but not "Archived" and vice-versa so separate theme settings for each would be handy.

"Draft" is only used for internal purposes when sending previews of content to others, whereas "Archived" could be for public content such as events that have passed.

anthonylindsay:
Isn't 'archived' an unpublished state?
Thus it too is internal.

Like, yeah, sure, we could have another setting for 'archived', but it seems somewhat overkill given both draft and archived are unpublished, no?

markconroy:
Archived is not necessarily unpublished. You could have an archive of past events for example, or might want to make content that is archived available to the public to see what you used to have. We've had councils who wish to make their archived content public, so I think making a separate checkbox for this would make sense.

anthonylindsay:
Right, ok I get what you're saying. So, it is possible to decouple the `archived` moderation state from being unpublished, but would you want the label on the node if it is published? Like, what would it actually mean?

Archiving means different things to different folks. You know yourself where one person will think archived == unpublished, another will think it's 'a different _view_ of old stuff', and another will think archiving  means removing the content from the site and putting it elsewhere.

As it's currently implemented, we're only dealing with unpublished nodes, so no harm, no foul.

Does adding the label make sense in a published scenario? Has anyone asked for that?

markconroy:
> Does adding the label make sense in a published scenario? Has anyone asked for that?

I think you're the first that has asked for any 'Archived' flag to be added to these things. Doing the decoupling from 'Draft' means "whatever" we want to do with it can more easily be done than forcing it on content if it's not desired even when 'Draft' is desired.

anthonylindsay:
Ok, sure. No big deal.

anthonylindsay:
OK, theme setting added, `archived` label decoupled from publication state.

localgov_base.theme
```diff
@@ -196,8 +196,20 @@ function localgov_base_preprocess_page(&$variables): void {
 function localgov_base_preprocess_node(&$variables): void {
   $node_status = $variables['node']->isPublished();
   if ($node_status === FALSE) {
-    if (theme_get_setting('localgov_base_add_draft_note_to_unpublished_content') === TRUE) {
-      $variables['label'] = t('[Draft]') . " " . $variables['node']->label();
+    if (theme_get_setting('localgov_base_add_draft_note_to_unpublished_content') == TRUE) {
```
markconroy: Let's put the `==` back to `===`

localgov_base.theme
```diff
@@ -196,8 +196,20 @@ function localgov_base_preprocess_page(&$variables): void {
...
+      $state = $variables['node']->get('moderation_state')->getString();
+      if ($state == 'draft') {
+        $variables['label'] = t('[Draft]') . " " . $variables['node']->label();
+      }
+    }
+  }
+  // We have a theme setting to prepend 'archived' to the title of
+  // archived content, but we don't mind whether it is published or not.
+  // An archived state does not have to always be unpubilshed.
+  if (theme_get_setting('localgov_base_add_archived_note_to_archived_content') === TRUE) {
```
markconroy, line 216: You will need to create a form item for this `localgov_base_add_archived_note_to_archived_content` setting/config

markconroy (changes requested):

localgov_base.theme
anthonylindsay: done!

localgov_base.theme
anthonylindsay: Gah! I had done, but had neglected to copy it over to the theme repo.
Workflow fail!

Is done now.

anthonylindsay:
Are there any other changes you'd like here?

markconroy:
I think the only thing missing was the /config/install entry. I've added that now, so will mark this as approved.

markconroy (approved):

PR #749, opened 2025-04-10
WIP: Add components to  node-full.html.twig

JohnAlbin:
DO NOT MERGE.

This is a work-in-progress proof-of-concept.

The PR will be recreated after governance has decided how to proceed on adding a design system (using Single Directory Components) to LocalGov.

### DIFF
--- a/css/base/variables.css
+++ b/css/base/variables.css
@@ -12,9 +12,9 @@ body {
   /* Widths and Spacing */
   --width-mega: 90rem; /* 1440px */
   --width-extra-large: 73.75rem; /* 1180px */
-  --width-large: 60rem;
-  --width-medium: 48rem;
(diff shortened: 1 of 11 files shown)


JohnAlbin:
The node--full.html.twig doesn't work in the PR at the moment, but this is roughly how I see it should look when adding components to LocalGov.

Note there is no HTML markup in this template file. Template files should add twig variables (content) to components in the design system.

https://github.com/localgovdrupal/localgov_base/blob/791b04bf2828989e8ed6f15c88f3b46633c212a8/templates/content/node--full.html.twig

markconroy:
Hi @JohnAlbin 

Just looking through this before Merge Tuesday later today. It looks really interesting. 

Are you planning on re-opening it/creating a different version of it?

PR #752, opened 2025-04-15
Add all font-awesome 6 icons

### LINKED ISSUE #696: Fontawesome - gather all icons into our base theme
@markconroy suggested we download all the regular fontawesome icons and include them in our localgov_base theme.

For example: https://github.com/FortAwesome/Font-Awesome/tree/6.x/svgs/regular

markconroy:
Closes #696 

## What does this change?

Adds every icon from Font Awesome 6 to our /icons folder so we can use them all and hopefully need to add fewer custom icons in custom themes.

## How to test

Add this to any template and it should load the AirBnB icon. Change airbnb to the name of some other icon if you want to test more. 

```
{% include "localgov_base:icon" with {
    icon_name: 'font-awesome/fa-6/svgs/brands/airbnb',
  }
%}
```

---


finnlewis:
Interesting! 

I guess we should just check with a few folks...

@andybroomfield thinks this is fine, as long as the licence is OK and filesize is reasonable. 

@Adnan-cds thinks it soudns good! 

@stephen-cox has no concerns... and mentions this could help us remove the fontawesome module which is only used for a single field on a single paragraph type!

finnlewis (approved):

PR #753, opened 2025-04-15
Add theme setting for JS breakpoint

### LINKED ISSUE #669: header.js presumes all councils want their mobile breakpoint to be 768px
We assumed for some reason that all councils would use 768. 

For custom themes, I think the easy thing for now is to duplicate the header.js, then use libraries-override to override the header-js library. But that's not scalable, and any council that does this will not get any updates that are added to header.js in the future.

We should try to make the breakbpoint value a variable, so we can use libraries extend instead of override. And then in our custom JS file, we'd just need something like 

```
import { header } from localgovbase...

let breakpoint = 960
```

markconroy:
Closes #669 

## What does this change?

Adds a theme setting for "JS mobile breakpoint" so different sites can have their JS mobile vs desktop at different widths. It defaults/fallsback to `768` which is our current value so no existing sites should be affected.

## How to test

Go to you custom theme settings if it's based on LocalGov Base, enter a numeric value in the JS mobile breakpoint field, check that the mobile menu works when your screen is below that value.

---

### DIFF
--- a/js/header.js
+++ b/js/header.js
@@ -2,9 +2,18 @@
  * @file JS file for the header component.
  */
 
-(function headerScript(Drupal) {
+(function headerScript(Drupal, drupalSettings, once) {
   Drupal.behaviors.header = {
     attach(context) {
+      let mobileBreakpointJS = 768;
+      // If there is a drupalSettings variable for the breakpoint, use that.
+      if (
+        drupalSettings.localgov_base &&
+        drupalSettings.localgov_base.mobileBreakpointJS
+      ) {
+        mobileBreakpointJS = drupalSettings.localgov_base.mobileBreakpointJS;
+      }
+
       // Hide the search form label. We use .once() to avoid re-running.
       //
       // @todo: make it possible to override this without having to maintain a
@@ -167,7 +176,7 @@
       function handleWindowResized() {
         handleReset();
 
-        if (window.innerWidth < 768) {
+        if (window.innerWidth < mobileBreakpointJS) {
           if (
             Object.keys(navInfo).includes('secondary') &&
             navInfo.secondary.toggle
@@ -244,4 +253,4 @@
       );
     },
   };
-})(Drupal);
+})(Drupal, drupalSettings, once);
--- a/js/header.js
+++ b/js/header.js
@@ -5,25 +5,7 @@
 (function headerScript(Drupal, drupalSettings, once) {
   Drupal.behaviors.header = {
     attach(context) {
-      let mobileBreakpointJS = 768;
-      // If there is a drupalSettings variable for the breakpoint, use that.
-      if (
-        drupalSettings.localgov_base &&
-        drupalSettings.localgov_base.mobileBreakpointJS
-      ) {
-        let mobileBreakpoint = drupalSettings.localgov_base.mobileBreakpointJS;
-
-        // 1. If it has 'px' at the end, remove it.
-        mobileBreakpoint = mobileBreakpoint.replace('px', '');
-
-        // 2. Convert it to a number.
-        mobileBreakpoint = parseInt(mobileBreakpoint, 10);
-
-        // 3. Check that the value is greater than 0.
-        if (mobileBreakpoint > 0) {
-          mobileBreakpointJS = mobileBreakpoint;
-        }
-      }
+      const mobileBreakpointJS = drupalSettings.localgov_base.mobileBreakpointJS;
 
       // Hide the search form label. We use .once() to avoid re-running.
       //
--- a/js/header.js
+++ b/js/header.js
@@ -11,7 +11,18 @@
         drupalSettings.localgov_base &&
         drupalSettings.localgov_base.mobileBreakpointJS
       ) {
-        mobileBreakpointJS = drupalSettings.localgov_base.mobileBreakpointJS;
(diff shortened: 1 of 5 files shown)


localgov_base.theme
```diff
@@ -87,12 +87,29 @@ function localgov_base_form_system_theme_settings_alter(&$form, FormStateInterfa
     '#default_value' => theme_get_setting('localgov_base_localgov_guides_vertical_navigation'),
     '#description'   => t('This will display the guide navigation vertically above the guide.'),
   ];
+
+  $form['localgov_base_mobile_breakpoint_js'] = [
+    '#type'          => 'textfield',
```
ctorgalson: This could be [a number field](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21Number.php/class/Number/10), e.g.:

```suggestion
    '#type'          => 'number',
```

It would allow the breakpoint determination (whether done in php or in javascript) to be slightly simpler.

localgov_base.theme
```diff
@@ -87,12 +87,29 @@ function localgov_base_form_system_theme_settings_alter(&$form, FormStateInterfa
...
 
 /**
  * Implements hook_preprocess_html().
  */
 function localgov_base_preprocess_html(&$variables): void {
+  // Add the mobile breakpoint value to drupalSettings for use in JavaScript.
+  if (!empty(theme_get_setting('localgov_base_mobile_breakpoint_js'))) {
+    $mobile_breakpoint = theme_get_setting('localgov_base_mobile_breakpoint_js');
+    $variables['#attached']['drupalSettings']['localgov_base']['mobileBreakpointJS'] = $mobile_breakpoint;
+  }
```
ctorgalson, line 123: if added along with changing the field type, and validation, this could eliminate the javascript checks since there'd _always_ be a value (and the other steps would guarantee it was numeric).

```suggestion
  }
  else {
    $variables['#attached']['drupalSettings']['localgov_base']['mobileBreakpointJS'] = 768;
  }
```

localgov_base.theme
```diff
@@ -87,12 +87,29 @@ function localgov_base_form_system_theme_settings_alter(&$form, FormStateInterfa
...
+
+  $form['localgov_base_mobile_breakpoint_js'] = [
+    '#type'          => 'textfield',
+    '#title'         => t('Mobile breakpoint for JavaScript'),
+    '#default_value' => theme_get_setting('localgov_base_mobile_breakpoint_js'),
+    '#description'   => t('This is the mobile breakpoint in pixels. This is used by the JS to determine when to show the mobile menu. If left blank, this will default to 768px.'),
+    '#attributes' => [
+      'placeholder' => '768',
+    ],
+
```
ctorgalson: I'd be tempted to add a validation function here, like 

```php
$form['#validate'][] = 'localgov_base_theme_settings_validate';
```

In concert with making the setting a number field, this could eliminate the need to handle the determination of the value in javascript since the value would be _guaranteed_ numeric (e.g. the validation could test `is_numeric()` to pass or fail validation.

js/subsites-menu.js
```diff
@@ -9,6 +9,26 @@
...
+        mobileBreakpoint = mobileBreakpoint.replace('px', '');
+
+        // 2. Convert it to a number.
+        mobileBreakpoint = parseInt(mobileBreakpoint, 10);
+
+        // 3. Check that the value is greater than 0.
+        if (mobileBreakpoint > 0) {
+          mobileBreakpointJS = mobileBreakpoint;
+        }
+      }
```
ctorgalson: I think that most/all of this could be handled in the existing `.theme` file functions (see below).

js/header.js
```diff
@@ -2,9 +2,29 @@
...
+
+        // 2. Convert it to a number.
+        mobileBreakpoint = parseInt(mobileBreakpoint, 10);
+
+        // 3. Check that the value is greater than 0.
+        if (mobileBreakpoint > 0) {
+          mobileBreakpointJS = mobileBreakpoint;
+        }
+      }
+
```
ctorgalson, line 9: I think almost all of this could be handled in the existing `.theme` file (see below). By forcing the input to be numeric, _validating_ that it _is_ numeric, and adding a default value in the theme file, I think we could reduce this to e.g. `const mobileBreakpointJS = drupalSettings?.localgov_base?.mobileBreakpointJS ?? 768;`

(Even adding the 768 in js is probably overkill since it'd only be necessary when something fairly catastrophic had happened to `drupalSettings` or the `.theme` file).

ctorgalson (approved):
As far as I can tell this will work perfectly well as is, so :+1:.

On the other hand, the few changes I suggested could take the (small!) extra load off of the javascript file, _and_ allow other Drupal behaviors to make use of the value without having to implement their own checks.

localgov_base.theme
markconroy: Validation added.

js/subsites-menu.js
markconroy: Done.

js/header.js
markconroy: Done

markconroy:
@ctorgalson Will you give this another require and approve/reject after my latest updates? Thanks.

ctorgalson (approved):
Looks great :tada: 

I think this will be quite useful.

PR #763, opened 2025-04-18
Add node-type class back

### LINKED ISSUE #762: Missing node--type-content-type class
After we moved node--full.html.twig to SDC we are now missing the `node--type-[content-type]` class. Let's add it back in in case anyone was using it.

markconroy:
Closes #762


finnlewis (approved):
Looks good to me!

PR #764, opened 2025-04-22
Set a variable for unpublished background colour

### LINKED ISSUE #754: Set the unpublished background colour as a variable
https://github.com/localgovdrupal/localgov_base/blob/1.x/css/components/unpublished-bg.css

Let's set this as a css variable so we don't have to override / extend the library if we want to change the colour.

markconroy:
Closes #754


msayoung (approved):
thanks!

PR #768, opened 2025-04-23
Add custom attributes for prev-next component

### LINKED ISSUE #758: Calling prev-next SDC from inside a node template gives duplicate classes
In our node template we have 

```
{% set classes = [
  'node',
  'more classes here'
]%}
```

In our prev-next component we have

```
{% set classes = [
  'lgd-prev-next',
  'more classes here'
]%}
```

If we create a node template that embeds/includes the prev-next SDC for the prev-next classes we get this:

```
<div class="node node--type-publications node--view-mode-teaser lgd-prev-next lgd-prev-news--publications"></div>
```

it's because we have `classes` as the same attribute name for the classes in both. 

I think we could fix this by creating a new attribute for prevNextClasses in the prev-next component.

```
{% set prevNextAttributes = create_attribute() %}
{% set prevNextClasses = [
  'lgd-prev-next',
  'lgd-pre-next--' ~. type,
]%}

<div{{ prevNextAttributes.addClass(prevNextClasses) }}>...</div>
```

markconroy:
Closes #758 

## What does this change?

Adds a custom attribute for prev-next component, so other attributes (like node classes) don't pollute it.

## How to test

Embed the prev-next component in a node template. Check that the classes are:

```
<nav class="node node--type-page lgd-prev-next">...</nav>
```

Now, check out this branch and check the classes now DO NOT have the node classes in the `nav`:

```
<nav class="lgd-prev-next">...</nav>
```

### DIFF
--- a/components/prev-next/prev-next.twig
+++ b/components/prev-next/prev-next.twig
@@ -7,6 +7,8 @@
  */
 #}
 
+{% set prevNextAttributes = prevNextAttributes|default(create_attribute) %}
+
 {% set icon_path = icon_path|default('@localgov_base/includes/icons') %}
 {% set prev_icon = prev_icon|default('chevron-left') %}
 {% set next_icon = next_icon|default('chevron-right') %}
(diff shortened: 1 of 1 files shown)


ctorgalson (approved):
Looks right to me :+1:

Issue #748, opened 2025-04-09
Dialog CSS conflict may cause styling anomalies


NikLP:
During the course of recent development I came across a strange problem where styles seemed to be disappearing from my webform dialog. Specifically a feedback form modal, which "lost" some styles as per pic: 

![Image](https://github.com/user-attachments/assets/e9982294-4ae1-4af5-97fe-e83e961843ff)

As it transpires, because of a potentially over-zealous tack adopted by the Gin theme (and potentially also Claro), css libraries for dialogs are being loaded where one might not expect. 

How/when this manifests:
- Logged out users may report missing dialog styles
- Logged in users may see dialog styles that are missing for logged out users

Why this is occurring:
When logged in users are using the Gin theme and the site theme, ie working with front-end pages that include the toolbar, dialog libraries are loaded and aggregated into the site css. When logged out users are delivered this aggregated css, they don't have access to the toolbar but they are getting css classes that don't exist for them, which override the default dialog classes.

Resolution:
- It is discussed in the thread below that it might be "fixed" upstream in gin and/or claro, but seems to be regarded as a subjective or case-by-case issue
- Recommendation is to omit loading of the libraries in question in one's own theme

https://www.drupal.org/project/gin_toolbar/issues/3181937

What that looks like in a nutshell:
Add the following lines to [frontend-theme].info.yml

```
libraries-override:
  claro/claro.drupal.dialog: false
  gin/dialog: false
  gin/gin_dialog: false (fixed for me without this line, needs assessment)
```

markconroy:
I'd like to keep using what Claro/Gin give us for the modal styling so we don't have to maintain it ourselves.

I wonder, instead of removing those libraries from our theme, could we do the opposite and make the dependencies in our theme so we always have the CSS needed?

@NikLP I can't reliably replicate this issue. Would you be able to do something like `hook_preprocess_page` in `localgov_base.theme` to check if 
- `gin_toolbar` exists and is installed, and if so to attach the necessary library from that module to all pages on the site
- `gin` theme exists and is enabled, and if so to attach the necessary library from that module to all pages on the site
- `claro` theme is enabled, and if so to attach the necessary library from that module to all pages on the site

I think that might be an interesting way to make sure we have the CSS we need without doing extra work ourselves.

NikLP:
@markconroy I understand that randomly removing libraries isn't the optimal solution, but:

- the dialogs look fine without them in the LGD theme
- the dialogs are (all but) broken with them included
[...]

PR #770, opened 2025-04-25
Override (remove) dialog libraries from backend themes to prevent CSS…

NikLP:
… issues in frontend themes

## What does this change?

Removes libraries from inclusion in frontend themes, to prevent a styling problem with dialogs which occurs when frontend themes and admin themes are present on the same page. This is incurred by aggregation only. Aggregated CSS files are polluted when admin users use the site.

CSS no longer broken and no other areas affected

Not 100% clear on the actual use case of the original inclusion of the libraries in question tbh!

## Images

See those on https://github.com/localgovdrupal/localgov_base/issues/748


markconroy (approved):
This looks "good enough for now, safe enough to try".

Let's approve.

PR #648, opened 2024-11-13
adds CSS to allow tables to scroll on small screens

markconroy:
Closes #579 

## What does this change?

1. Sets `.layout__region` and `.field--table` to `max-width: 100%`, so they don't overflow with any tables (or other elements)
2. Sets the .tablefield-wrapper to use `overflow-x: auto` so if it's too wide, it will have a horizonal scroll.

## How to test

1. Visit /test-subsite-demo-all-components/example-page-tables-and-tabs on a small screen and check that only the table has horizontal scroll, not the whole page.

---
Thanks to [Big Blue Door](https://www.bigbluedoor.net/) for sponsoring my time to work on this.

PR #771, opened 2025-05-14
Add JS for responsive tables in long text fields

### LINKED ISSUE #759: Table added through the WYSIWYG breaks content reflow
If we add a table through the WYSIWYG (of three columns or more), and view the site at 320px wide, we are forced to use a horizontal scroll. 

The WCAG criterion Reflow https://www.w3.org/WAI/WCAG21/Understanding/reflow.html 

> Content can be presented without loss of information or functionality, and without requiring scrolling in two dimensions for:
> 
> Vertical scrolling content at a width equivalent to 320 [CSS pixels](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html#dfn-css-pixel);
> Horizontal scrolling content at a height equivalent to 256 [CSS pixels](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html#dfn-css-pixel).
> Except for parts of the content which require two-dimensional layout for usage or meaning.

If we can put an overflow-x: scroll on the table i think this would solve it.

markconroy:
Closes #759 

## What does this change?

- Finds tables in long text fields, clones the content of the table, and then places it inside a div with a class of `lgd-responsive-table`, so we can then use `overflow-x: auto` on these containers.
- Adds a theme setting - turned on by default for new installs - to allow existing sites use this feature

## How to test

- Go to your theme settings page, check the checkbox for 'Make WYSIWYG tables responsive'
- Create a table in a long text field
- Reduce the size of your screen to mobile size
- Check that the table can be scrolled horizontally, but the page does not scroll with it

## Caveat

Do we want to use a theme setting to enable this?

## Related PR
https://github.com/localgovdrupal/localgov_base/pull/648

^ That PR is for the table field tables rather than wysiwyg tablets iirc.

---

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -99,6 +99,14 @@ function localgov_base_form_system_theme_settings_alter(&$form, FormStateInterfa
     ],
   ];
 
+  $form['localgov_base_responsive_wysiwyg_tables'] = [
+    '#type'          => 'checkbox',
+    '#title'         => t('Make WYSIWYG tables responsive.'),
+    '#default_value' => theme_get_setting('localgov_base_responsive_wysiwyg_tables'),
+    '#description'   => t('Places tables inserted via WYSIWYG into a responsive container, so they can scroll horizontally on small screens.'),
+    '#default_value' => theme_get_setting('localgov_base_responsive_wysiwyg_tables') ? theme_get_setting('localgov_base_responsive_wysiwyg_tables') : FALSE,
+  ];
+
   $form['#validate'][] = 'localgov_base_theme_settings_validate';
 }
 
@@ -145,6 +153,13 @@ function localgov_base_preprocess_html(&$variables): void {
       if (theme_get_setting('localgov_base_header_behaviour') === 'appears_on_scroll') {
         $variables['attributes']['class'][] = 'sticky-header--scroll';
       }
+    }
+  }
+
+  // Add the 'localgov_base_responsive_wysiwyg_tables' value to drupalSettings.
+  if (!empty(theme_get_setting('localgov_base_responsive_wysiwyg_tables'))) {
+    if (theme_get_setting('localgov_base_responsive_wysiwyg_tables') === TRUE || 1) {
+      $variables['#attached']['library'][] = 'localgov_base/responsive-tables';
     }
   }
 }
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -102,7 +102,6 @@ function localgov_base_form_system_theme_settings_alter(&$form, FormStateInterfa
   $form['localgov_base_responsive_wysiwyg_tables'] = [
     '#type'          => 'checkbox',
     '#title'         => t('Make WYSIWYG tables responsive.'),
-    '#default_value' => theme_get_setting('localgov_base_responsive_wysiwyg_tables'),
     '#description'   => t('Places tables inserted via WYSIWYG into a responsive container, so they can scroll horizontally on small screens.'),
     '#default_value' => theme_get_setting('localgov_base_responsive_wysiwyg_tables') ? theme_get_setting('localgov_base_responsive_wysiwyg_tables') : FALSE,
   ];
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - '1.x'
+      - '2.x'
   pull_request:
     branches:
       - '1.x'
+      - '2.x'
   workflow_dispatch:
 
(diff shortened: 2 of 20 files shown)


willguv:
@markconroy this feels like something could add to the product newsletter too. Do councils just update localgov_base?

Thanks for doing this

msayoung:
Hey @markconroy - this works fine. 

I'm just wondering if we should use something like this instead ? 
https://www.drupal.org/project/responsive_tables_filter which allows for some options on how to manage responsive tables. 

What do you think?

markconroy:
I wasn't aware of that module. Let's chat about it in MT. 

I have a feeling wiht that module, it's changing the design/layout of tables pretty drastically, so maybe it would be worth putting that module in "as well as" this PR. Then sites can decide if they want the full-on experience of that module, or something similar like this PR per site.

msayoung:
I won't make it to MT today, dentist calls. 

This module can be set to default to the same behaviour you've made, with the option to override it on a case by case basis. That might be the safest option. 

The downside is that it only works if there's a table header set, which there should be anyway. 

There is also this module https://www.drupal.org/project/responsive_table_filter which seems to do the same as yours, if we want to stick to contrib ?  I haven't looked at how it does it though.

msayoung (approved):
Discussed with Mark and decided to go with this approach for now, and people can not enable this and use whichever other module they like.

finnlewis:
We want this to merge into the 2.x branch first, then backport to the 1.x.

So maybe we recreate the merge requests branching off the 2.x branch.

Or resolve the merge conflicts.

@tonypaulbarker will take a look :)

localgov_base.theme
```diff
@@ -99,6 +99,14 @@ function localgov_base_form_system_theme_settings_alter(&$form, FormStateInterfa
     ],
   ];
 
+  $form['localgov_base_responsive_wysiwyg_tables'] = [
+    '#type'          => 'checkbox',
+    '#title'         => t('Make WYSIWYG tables responsive.'),
+    '#default_value' => theme_get_setting('localgov_base_responsive_wysiwyg_tables'),
+    '#description'   => t('Places tables inserted via WYSIWYG into a responsive container, so they can scroll horizontally on small screens.'),
+    '#default_value' => theme_get_setting('localgov_base_responsive_wysiwyg_tables') ? theme_get_setting('localgov_base_responsive_wysiwyg_tables') : FALSE,
```
millnut, line 106: Duplicate `#default_value` with line 105 above

millnut (changes requested):
Also looks like this PR has conflicts

tonypaulbarker:
Some conflicts in grid.css 

I assume net changes to this file should be zero.

I think that we want a grid of '12' columns, but just to be sure:

```
.lgd-row {
  display: grid;
  gap: var(--grid-column-spacing);
  grid-template-columns: repeat(12, 1fr);
}
```
or 
```
.lgd-row {
  display: grid;
  gap: var(--grid-column-spacing);
  grid-template-columns: 1fr;
}
```

```
.lgd-row__one-quarter,
.lgd-row--quarters > *,
.lgd-row__one-third,
.lgd-row--thirds > *,
.lgd-row__one-half,
.lgd-row--halves > *,
.lgd-row__two-thirds,
.lgd-row__three-quarters,
.lgd-row__full {
  grid-column: span 12;
  width: 100%;
}
```
or 
```
.lgd-row__one-quarter,
.lgd-row--quarters > *,
.lgd-row__one-third,
.lgd-row--thirds > *,
.lgd-row__one-half,
.lgd-row--halves > *,
.lgd-row__two-thirds,
.lgd-row__three-quarters,
.lgd-row__full {
  width: 100%;
  grid-column: span 1;
}
```

localgov_base.theme
tonypaulbarker: @markconroy @millnut  line 107 has some additional logic compared to line 105. Presume we want line 107. "If we have this setting, use the setting, otherwise false" ?

tonypaulbarker:
Conflicts should be well now. Please review comments, review code and re-test this works as expected.

msayoung:
Seems to work fine @tonypaulbarker , but the are some failing tests. Can you have a look at those?

markconroy:
The failing tests are:

1. YAML - belonging to the 'display full' SDC, we should file a follow up issue to fix that.
2. PHP Unit - geolocation module
3. PHP Unit - localgov_paragraphs module
4. Stylelint - false positive

@millnut is the changes you requested sorted now? If so, I think this is ready to merge.

finnlewis:
Discussed in Merge Tuesday and all happy to merge and release.

PR #790, opened 2025-05-28
Wrap icon and text in <a> tag

### LINKED ISSUE #787: Link paragraph (button style) : whole element should be  clickable
The link paragraph: https://github.com/localgovdrupal/localgov_base/blob/1.x/templates/paragraphs/paragraph--localgov-link.html.twig

This contains an icon and the link text. The whole thing changes state on hover, however the link only wraps the text , not the icon. 

See: 

https://site2.microsites.localgovdrupal.org/

If you hover over the Email us link icon, it appears to be clickable , but only the text is.

markconroy:
Closes #787 

## What does this change?

Moves the icon in the link paragraph into the `a` tag so both the text and the icon are clickable.

### DIFF
--- a/css/components/link.css
+++ b/css/components/link.css
@@ -13,6 +13,7 @@
 }
 
 .link--button-style .link-wrapper {
+  position: relative;
   padding: var(--button-padding-vertical) var(--button-padding-horizontal) var(--button-padding-vertical) var(--button-padding-horizontal);
   color: var(--button-text-color);
   border: var(--border);
@@ -36,7 +37,11 @@
   text-decoration: none;
   color: inherit;
 }
-
+.link--button-style a::after {
+  content: "";
+  position: absolute;
+  inset: 0;
+}
 .link--button-style:focus .link-wrapper,
 .link--button-style:hover .link-wrapper {
   text-decoration: underline;
--- a/css/components/link.css
+++ b/css/components/link.css
@@ -3,6 +3,10 @@
   align-items: center;
 }
 
+.link-content > a {
+  display: flex;
+}
+
 .link-icon {
   flex-shrink: 0;
   margin-inline-end: var(--link-icon-margin);
--- a/css/components/link.css
+++ b/css/components/link.css
@@ -9,9 +9,9 @@
 }
 
 .link-content > a::after {
-  content: "";
   position: absolute;
   inset: 0;
+  content: "";
 }
 
 .link-icon {
--- a/css/components/link.css
+++ b/css/components/link.css
@@ -1,4 +1,5 @@
 .link-wrapper {
+  position: relative;
   display: flex;
   align-items: center;
 }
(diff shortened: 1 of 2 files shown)


msayoung (changes requested):
This works fine for the default styling, but the "button" styling still has an area which looks clickable (and changes on hover) but isn't clickable.  

image

markconroy:
@msayoung Can you test this again please? It looks to me that the template only renders if there is a URL (without a URL, it's basically just a text field).

I've added:

1. Link with a URL - text + icon is clickable
2. Button with a URL - text + icon is clickable
3. Link without a URL - nothing displays on page from template
4. Button without a URL - nothing displays on page from template

Here's a screenshot of the edit page (you can see the two extra items) followed by a screenshot of the rendered page (only the items with URLs are showing). So, _I think_ this is working correctly in terms of "icon + text should be clickable".

Screenshot 2025-06-24 at 12 32 28


Screenshot 2025-06-24 at 12 31 14

msayoung:
@markconroy - I think we are looking at different things.

 The issue is that with the button style, there is a hover effect on the non-interactive element. 

```
<div class="link link--button-style paragraph paragraph--type--localgov-link paragraph--view-mode--default">
      <div class="link-wrapper">
      <div class="link-content">
        <a href="https://agile.coop">
        <div aria-hidden="true" class="lgd-icon link-icon">
          <svg xmlns="http://www.w3.org/2000/svg" focusable="false" viewBox="0 0 320 512">...
        </svg>
        </div> This is a link
        </a>
      </div>
    </div>
</div>
```

The hover effect is on the .link-wrapper , which is much larger than the <a> , making it look like the whole thing is clickable.

Does that make sense?

markconroy:
Thanks @msayoung I see what you mean now.

I've added some CSS so the whole component is clickable.

css/components/link.css
```diff
@@ -1,8 +1,19 @@
 .link-wrapper {
+  position: relative;
   display: flex;
   align-items: center;
```
ctorgalson, line 4: If the only content of `.link-wrapper` is `.link-content`, I think line 3 is superfluous, and line 4 belongs after line 8.

ctorgalson (commented):
No objections, just one small suggestion. I can see that this should work as intended.

I *do* wonder about the need for `.link-content > a::after` though--it would be simple enough to shift the styles from `.link-wrapper` into `.link-content > a` so that the link itself occupied 100% of `.link-wrapper`.

Is it just a cautious approach to lessen the risk of stepping on the toes of child-theme implementations?

markconroy:
@ctorgalson 
> I do wonder about the need for .link-content > a::after though--it would be simple enough to shift the styles from .link-wrapper into .link-content > a so that the link itself occupied 100% of .link-wrapper.

I think I'd like to leave this as is "just in case" someone wants the link to be 100% of the .link-content (which in our default set up is the same as 100% of .link-wrapper) but might also want some padding/different colour, etc on .link-wrapper.

markconroy:
@msayoung I think i have your issues resolved, but it still says "Changes requested" by you. Can you hit the approve button if you don't have any more changes requested please.

msayoung (approved):

PR #797, opened 2025-05-30
fix: missing prev-next attributes for nav

millnut:
## What does this change?

Recent changes in prev-next have broken tests, see https://github.com/localgovdrupal/localgov/actions/runs/15327716159/job/43126523978

This is because the classes are missing on the `<nav>` element

With this PR it reinstates these to look like (publications as an example);
```
<nav class="lgd-prev-next lgd-prev-next--publications">
```

Currently these classes are missing and this is shown
```
<nav>
```

### DIFF
--- a/components/prev-next/prev-next.twig
+++ b/components/prev-next/prev-next.twig
@@ -7,7 +7,7 @@
  */
 #}
 
-{% set prevNextAttributes = prevNextAttributes|default(create_attribute) %}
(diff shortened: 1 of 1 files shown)


finnlewis:
Thanks @millnut  - any idea why the workflow is not running?

msayoung (approved):

millnut:
> Thanks @millnut - any idea why the workflow is not running?

Because of this https://github.com/localgovdrupal/localgov_base/pull/799 would be my guess, looks like they are running now

PR #803, opened 2025-06-05
Add default background colour for callouts

### LINKED ISSUE #802: Add default background colour for .callout
Our `.callout` class has the text set to white by default. 

If we remove some callout types (e.g. teal, carbon) from our wysiwyg styles, then anything that had been using those will now be white text on white background.

It also means if you set a default background colour in a subtheme, it overrides all the other background colours for callouts, since custom background colour comes after localgov bases's css.

We should set a default background colour (probably accent colour) to `.callout`.

markconroy:
Closes #802

### DIFF
--- a/css/components/callout.css
+++ b/css/components/callout.css
@@ -5,6 +5,7 @@
   padding: 1.5rem;
   text-align: center;
   color: var(--color-white);
+  background-color: var(--color-accent);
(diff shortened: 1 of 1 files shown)


msayoung (approved):
Works. Need to update the ckeditor styles also. will create another issue.

Issue #805, opened 2025-06-12
Add default accessible label to templates using localgov_base:prev-next as needed


ctorgalson:
Currently `localgov_base` lacks these--IIRC, we removed one from one of the consuming templates when componentizing the Prev-Next nav, but didn't add it back. We can add them by default.

There are recent fixes to the component's `prevNextAttributes` variable that can enable us to add e.g. `aria-label` and `aria-labelledby`. Using `aria-label` is thus possible immediately, but following Drupal's usual `<nav>` convention is not currently possible.

tonypaulbarker:
These changes were merged. Closing.

PR #806, opened 2025-06-12
Feature/2.x/805 add default accessible label to templates using localgov prev next

ctorgalson:
Re: #805

## What does this change?

This makes several changes:

- documents existing but undocumented prop `prevNextAttributes` in `prev-next.component.yml`
- adds new `prev_next_nav_aria_label` prop for convenience (i.e. since this can also be set using `prevNextAttributes`)
- adds new `prev_next_pre_content` slot to add content before the Prev and Next links
- adds new `prev_next_post_content` slot to add content after the Prev and Next links
- uses `prev_next_pre_content` slot to add a labelled, visually-hidden heading to Blog template
- uses `prev_next_pre_content` slot to add a labelled, visually-hidden heading to Guides template
- uses `prev_next_pre_content` slot to add a labelled, visually-hidden heading to Publication template
- uses `prev_next_pre_content` slot to add a labelled, visually-hidden heading to Step by Step template

The new labels all use Drupal's stock pattern: an `aria-labelledby` attribute on the `<nav>` element referring to a visually-hidden child `<h2>` with a matching `id` attribute.

## How to test

Test all of the following with demo content (not sure what module the Blog template requires):

- /publications/publications-cover-page-demo-content/petitions-scheme-publications-demo-content-1
  - browser: Prev-Next heading not visible
  - screenreader: h2 'Publication navigation' announced
- /order-copy-certificate/order-copy-certificate
  - browser: Prev-Next heading not visible
  - screenreader: h2 'Guides navigation' announced
- /adult-health-and-social-care/step-by-step/request-support-adult-step-step/choosing-how-manage-your
  - browser: Prev-Next heading not visible
  - screenreader: h2 'Step by step navigation' announced

- the various `<nav>` elements wrapping the Prev-Next links will now have an accessible text label by default

For existing overrides of the template, a new `aria-labelledby` and `h2` will appear:

- if the theme template uses the SDC and has _not_ provided its own accessible label, the component will now have one; this is an improvement
- if the theme template uses the SDC _has_ provided its own label using the only currently available method (`aria-label`), [the `<nav>`'s `aria-labelledby` attribute plus its new label will override the theme's label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-labelledby#benefits_and_drawbacks); this is _probably_ not serious but could be a problem where e.g. the new label string is not translated on a multilingual site

## Images

n/a

## Accessibility

-   [ ] ~[Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)~ 
-   [ ] ~[The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)~

### DIFF
--- a/templates/block/guides-prev-next-block.html.twig
+++ b/templates/block/guides-prev-next-block.html.twig
@@ -1,4 +1,4 @@
-{% set nav_id = 'guides-prev-next' %}
+{% set nav_id = 'guides-prev-next-' ~ random() %}
 {% set prevNextAttributes = create_attribute({ id: nav_id }) %}
 {% set prev_next_type = 'guides' %}
 {#
--- a/templates/block/guides-prev-next-block.html.twig
+++ b/templates/block/guides-prev-next-block.html.twig
@@ -1,3 +1,5 @@
+{% set nav_id = 'guides-prev-next' %}
+{% set prevNextAttributes = create_attribute({ id: nav_id }) %}
 {% set prev_next_type = 'guides' %}
 {#
   Probably the parent module should take responsibility for this.
@@ -9,4 +11,8 @@
 {% set prev_url = previous_url|default('') %}
 {% set prev_title = previous_title %}
 
-{% include 'localgov_base:prev-next' %}
+{% embed 'localgov_base:prev-next' %}
+  {% block prev_next_pre_content %}
+    <h2 class="visually-hidden" aria-labelledby="{{ nav_id }}">{{ 'Guides navigation'|t }}</h2>
+  {% endblock %}
+{% endembed %}
--- a/templates/block/localgov-blogs-prev-next-block.html.twig
+++ b/templates/block/localgov-blogs-prev-next-block.html.twig
@@ -1,2 +1,8 @@
+{% set nav_id = 'blogs-prev-next' %}
+{% set prevNextAttributes = create_attribute({ id: nav_id }) %}
 {% set prev_next_type = 'blog' %}
-{% include 'localgov_base:prev-next' %}
+{% embed 'localgov_base:prev-next' %}
+  {% block prev_next_pre_content %}
+    <h2 class="visually-hidden" aria-labelledby="{{ nav_id }}">{{ 'Blog navigation'|t }}</h2>
+  {% endblock %}
+{% endembed %}
--- a/templates/block/localgov-blogs-prev-next-block.html.twig
+++ b/templates/block/localgov-blogs-prev-next-block.html.twig
@@ -1,4 +1,4 @@
-{% set nav_id = 'blogs-prev-next' %}
+{% set nav_id = 'blogs-prev-next-' ~ random() %}
 {% set prevNextAttributes = create_attribute({ id: nav_id }) %}
 {% set prev_next_type = 'blog' %}
 {% embed 'localgov_base:prev-next' %}
--- a/templates/navigation/book-navigation--publication.html.twig
+++ b/templates/navigation/book-navigation--publication.html.twig
@@ -32,9 +32,15 @@
 #}
 
 {% if has_links %}
+  {% set nav_id = 'publication-prev-next' %}
+  {% set prevNextAttributes = create_attribute({ id: nav_id }) %}
   {# This uses previous_* vars, prev-next uses prev_*, so we translate. #}
   {% set show_title = true %}
   {% set prev_next_type = 'publications' %}
 
-  {% include 'localgov_base:prev-next' %}
+  {% embed 'localgov_base:prev-next' %}
+    {% block prev_next_pre_content %}
+      <h2 class="visually-hidden" aria-labelledby="{{ nav_id }}">{{ 'Publication navigation'|t }}</h2>
+    {% endblock %}
+  {% endembed %}
 {% endif %}
--- a/templates/navigation/book-navigation--publication.html.twig
+++ b/templates/navigation/book-navigation--publication.html.twig
@@ -32,7 +32,7 @@
 #}
 
 {% if has_links %}
-  {% set nav_id = 'publication-prev-next' %}
+  {% set nav_id = 'publication-prev-next-' ~ random() %}
   {% set prevNextAttributes = create_attribute({ id: nav_id }) %}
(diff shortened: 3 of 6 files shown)


tonypaulbarker:
We have some test failures that we need to check out. 

Let's add to - "Have we considered potential risks?"
Will this break for someone who has overridden any of those templates and still includes the SDC?
Is there ever a case where we have two of these on a page and we have to consider duplicate IDs?

I _think_ the answer to both of these will be 'no'.

templates/block/guides-prev-next-block.html.twig
```diff
@@ -9,4 +11,8 @@
 {% set prev_url = previous_url|default('') %}
 {% set prev_title = previous_title %}
 
-{% include 'localgov_base:prev-next' %}
+{% embed 'localgov_base:prev-next' %}
+  {% block prev_next_pre_content %}
+    <h2 class="visually-hidden" aria-labelledby="{{ nav_id }}">{{ 'Guides navigation'|t }}</h2>
```
markconroy, line 16: Should we make a variable for 'Guides navigation' just in case someone wants to use something different for whatever reason (such as the name of the actual guide)?

templates/block/localgov-blogs-prev-next-block.html.twig
```diff
@@ -1,2 +1,8 @@
+{% set nav_id = 'blogs-prev-next' %}
+{% set prevNextAttributes = create_attribute({ id: nav_id }) %}
 {% set prev_next_type = 'blog' %}
-{% include 'localgov_base:prev-next' %}
+{% embed 'localgov_base:prev-next' %}
+  {% block prev_next_pre_content %}
+    <h2 class="visually-hidden" aria-labelledby="{{ nav_id }}">{{ 'Blog navigation'|t }}</h2>
```
markconroy, line 6: Should we make a variable for 'Blog navigation' just in case someone wants to use something different for whatever reason (such as the name of the blog - "Week notes blog navigation" / "Council Tax Blog navigation")?

templates/navigation/book-navigation--publication.html.twig
```diff
@@ -32,9 +32,15 @@
...
+  {% set nav_id = 'publication-prev-next' %}
+  {% set prevNextAttributes = create_attribute({ id: nav_id }) %}
   {# This uses previous_* vars, prev-next uses prev_*, so we translate. #}
   {% set show_title = true %}
   {% set prev_next_type = 'publications' %}
 
-  {% include 'localgov_base:prev-next' %}
+  {% embed 'localgov_base:prev-next' %}
+    {% block prev_next_pre_content %}
+      <h2 class="visually-hidden" aria-labelledby="{{ nav_id }}">{{ 'Publication navigation'|t }}</h2>
```
markconroy, line 43: Should we make a variable for 'Publication navigation' just in case someone wants to use something different for whatever reason (such as the name of the actual publication)?

templates/views/views-view-list--localgov-step-by-step-navigation--prev-next.html.twig
```diff
@@ -36,8 +36,14 @@
  */
 #}
 
+{% set nav_id = 'step-by-step-prev-next' %}
```
markconroy: We should probably stick a random id or something at the end of this, just to make sure we don't have duplicate ids.

templates/block/localgov-blogs-prev-next-block.html.twig
```diff
@@ -1,2 +1,8 @@
+{% set nav_id = 'blogs-prev-next' %}
```
markconroy: We should probably stick a random id or something at the end of this nav_id, just to make sure we don't have duplicate ids.

markconroy (changes requested):
This looks really good, just a few small comments to look at and we're good to go.

templates/block/guides-prev-next-block.html.twig
ctorgalson: @markconroy To override a variable like that, themes would have to implement this template, right? If they do that, they can redefine the existing `prev_next_pre_content` slot.

templates/block/localgov-blogs-prev-next-block.html.twig
ctorgalson: Will follow whatever we do for Guides here.

templates/navigation/book-navigation--publication.html.twig
ctorgalson: Will follow whatever we do for Guides here.

finnlewis:
@ctorgalson still wants to tweak something on this one before it's ready.

templates/views/views-view-list--localgov-step-by-step-navigation--prev-next.html.twig
ctorgalson: Added in b8353f44.

templates/block/localgov-blogs-prev-next-block.html.twig
ctorgalson: Added in b8353f44.

ctorgalson:
@markconroy I think b8353f44 resolves the issues here if you want to give it another once-over.

ctorgalson:
@tonypaulbarker 

> Let's add to - "Have we considered potential risks?"
> Will this break for someone who has overridden any of those templates and still includes the SDC?
> Is there ever a case where we have two of these on a page and we have to consider duplicate IDs?
>
> I think the answer to both of these will be 'no'.

Taking the second first, since you and @markconroy suggested the same issue, I've added a random string to those generated ids. It will make them unique at the cost of using them as selectors. But since they're _new_, that's a non-issue.

For the first point, I tried to think how the change could break a child-theme, but I couldn't come up with a way:

- the changes to the component itself consist of a new prop and two new slots, so those won't affect anything,
- the changes to the instances using the SDC probably can't cause issues because a child-theme that implements the same templates has _overridden_ them

But if anyone thinks of a way it could go wrong, I'm happy to try to address it in the code (and acknowledge the creativity!)

markconroy (approved):
LGTM

Issue #807, opened 2025-06-16
Invalid max-height value in secondary-menu.css


ctorgalson:
When overriding a `max-height` value for smaller screens, [the desktop styles in `secondary-menu.css` use `max-height: auto;`](https://github.com/localgovdrupal/localgov_base/blob/2.x/css/components/secondary-menu.css#L32). But [`auto` is not a possible value for `max-height`](https://developer.mozilla.org/en-US/docs/Web/CSS/max-height#values). This is definitely an edge-case, but there are times when it results in menu items being rendered outside of the main menu--they're still technically visible, but depending on the design may become totally inaccessible due to placement/contrast.

It's typically unset using `none` (the browser default), though there are lots of other options for specific cases. Here, I think, `none` will do nicely.

markconroy:
Let's close this if the corresponding PR has been merged.

@ctorgalson if you start your PR message with **Closes #123456** instead of **Re #123456** then the issue will automatically be closed when the PR is merted.

PR #808, opened 2025-06-16
fix: switches secondary nav max-height from 'auto' to 'none' at deskt…

ctorgalson:
Re: #807 

...op sizes

"auto" is not a possible value, so it fails to unset the max-height value set for mobile/tablet sizes, sometimes obscuring menu items.

## To reproduce

This is relatively difficult to trigger on the LGD demo because the menu only has three items. The following steps make it seem absurd, but with menus containing 20+ items, it's easier to trigger :)

1. load a page like e.g. https://demo.localgovdrupal.org/localgov-drupal-demo
2. in the browser's css rules panel, find the selector `.region-secondary-menu .menu-item > a`
3. to that selector, add `font-size: 5em;`

The result is like this:

image

## What does this change?

The new code uses a valid keyword for `max-height` in `secondary-nav.css`.

## How to test

Test as per "To reproduce" above

If the new desktop `max-height` value is honoured by browsers.

This comes with very few risks:

- in most cases, it will be undetectable
- in some cases (like the one that prompted this issue), it will _fix_ a problem
- in a very few cases, where a child theme's css is unintentionally (i.e. due to this bug) relying on the _mobile_ styles' `max-height` value _at desktop sizes_, along with something like `oveflow-y: scroll;`, this could cause menu items to exceed the bounds of the menu container; this would be correctable with a single css rule

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### DIFF
--- a/css/components/secondary-menu.css
+++ b/css/components/secondary-menu.css
@@ -29,7 +29,7 @@
 
   .region-secondary-menu .menu {
     overflow-y: initial;
-    max-height: auto;
+    max-height: none;
     column-count: var(--secondary-menu-column-count);
   }
 }


finnlewis (approved):
Looks good to me!

PR #812, opened 2025-06-24
Feature/811 once and context

### LINKED ISSUE #811: Use context + once instead of document + querySelector in subsites-menu.js
In an effort to avoid JS firing more than once, and to follow Drupal best practices.

markconroy:
Closes #811

### DIFF
--- a/js/subsites-menu.js
+++ b/js/subsites-menu.js
@@ -11,10 +11,8 @@
 
       const { mobileBreakpointJS } = drupalSettings.localgov_base;
 
-      const subsitesMenuToggle = document.querySelector(
-        '.subsite-extra__header-toggle-button',
-      );
-      const subsitesMenu = document.querySelector('.subsite-extra-menu');
+      const [subsitesMenuToggle] = once('allSubsitesMenuToggles', '.subsite-extra__header-toggle-button', context);
+      const [subsitesMenu] = once('allSubsitesMenuToggles', '.subsite-extra-menu', context);
 
       subsitesMenuToggle.addEventListener('click', () => {
         subsitesMenuToggle.setAttribute(
@@ -54,7 +52,7 @@
       });
 
       // Close the menu when a click is made outside of it.
-      document.addEventListener('click', (e) => {
+      context.addEventListener('click', (e) => {
         if (!e.target.closest('#lgd-header__nav--subsites-menu')) {
           handleReset();
         }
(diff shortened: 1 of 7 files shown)


js/subsites-menu.js
```diff
@@ -11,10 +11,8 @@
 
       const { mobileBreakpointJS } = drupalSettings.localgov_base;
 
-      const subsitesMenuToggle = document.querySelector(
-        '.subsite-extra__header-toggle-button',
-      );
-      const subsitesMenu = document.querySelector('.subsite-extra-menu');
+      const [subsitesMenuToggle] = once('allSubsitesMenuToggles', '.subsite-extra__header-toggle-button', context);
+      const [subsitesMenu] = once('allSubsitesMenuToggles', '.subsite-extra-menu', context);
```
ctorgalson, line 15:
```suggestion
      const [subsitesMenu] = once('allSubsitesMenuToggles', '.subsite-extra-menu', context);
      
      if (!subsitesMenuToggle || !subsitesMenu) {
        return;
      }
```

This is untested, but it's the usual defensive companion to `.once()`.

ctorgalson (commented):
Late review, sorry.

TL;DR: I think this is good but needs a guard clause to complete #811; I didn't have a chance to test it.

Using `.once()` will only _return_ DOM elements once, but I think the _behavior_ will still run multiple times. 

The second part we usually add is one or more guard clauses (see suggestion). This makes sure that, on subsequent runs, the behavior doesn't try to _do_ anything and exits immediately.

For example, if this behavior ran twice, the first go would mark the menu and toggle with a `data-once` attribute and continue through the code. But on the second iteration, I think the two variables will be undefined, and [line 17](https://github.com/localgovdrupal/localgov_base/blob/ab2b0582d10846235652d1a81c95a40a50850a42/js/subsites-menu.js#L17) will try to add an event listener to an undefined element.

js/subsites-menu.js
markconroy: Thanks @ctorgalson, I've added in the extra defensive clause now. Let me know if you think we're missing any thinbg.

ctorgalson (approved):

finnlewis:
For reference: https://www.drupal.org/docs/drupal-apis/javascript-api/javascript-api-overview#s-example-using-drupalbehaviors

PR #813, opened 2025-06-24
Move theme settings into fieldsets and containers

### LINKED ISSUE #791: Tidy up theme settings form
We have lots of settings on the theme settings page. It might be good to group them in to fieldsets/accordions or something like that.

markconroy:
Closes #791 

## What does this change?

- Adds a general container for the localgov_base settings.
- Adds fieldsets inside this container to group the different settings types

## How to test

Change some settings, make sure they take effect and other settings do not break.

We have a tidier settings for for localgov_base and subthemes of it.

Settings for current sites break (this should not happen).

## Images

Screenshot 2025-06-24 at 14 03 53

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -20,49 +20,29 @@ function localgov_base_form_system_theme_settings_alter(&$form, FormStateInterfa
     return;
   }
 
-  $form['localgov_base_remove_css'] = [
-    '#type'          => 'checkbox',
-    '#title'         => t('Remove CSS libraries from base theme.'),
-    '#default_value' => theme_get_setting('localgov_base_remove_css'),
-    '#description'   => t("Check this box to disable the base theme's CSS"),
+  // Add settings in a details element for better organization
+  $form['localgov_base_settings'] = [
+    '#type' => 'details',
+    '#title' => t('LocalGov Base Settings'),
+    '#open' => TRUE,
   ];
 
-  $form['localgov_base_remove_js'] = [
-    '#type'          => 'checkbox',
-    '#title'         => t('Remove JS libraries from base theme.'),
-    '#default_value' => theme_get_setting('localgov_base_remove_js'),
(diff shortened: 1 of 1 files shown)


localgov_base.theme
```diff
@@ -20,85 +20,121 @@ function localgov_base_form_system_theme_settings_alter(&$form, FormStateInterfa
     return;
   }
 
-  $form['localgov_base_remove_css'] = [
+  // Add settings in a details element for better organization
+  $form['localgov_base_settings'] = [
+    '#type' => 'details',
+    '#title' => t('LocalGov Base Settings'),
```
msayoung: This title get's inherited by the child themes. Should it change to the current theme name?

localgov_base.theme
```diff
@@ -20,85 +20,121 @@ function localgov_base_form_system_theme_settings_alter(&$form, FormStateInterfa
...
     '#title'         => t('Remove JS libraries from base theme.'),
     '#default_value' => theme_get_setting('localgov_base_remove_js'),
     '#description'   => t("Check this box to disable the base theme's JavaScript"),
   ];
 
-  $form['localgov_base_add_unpublished_background_colour'] = [
+  // Create a fieldset for content status related settings
+  $form['localgov_base_settings']['content_status_fieldset'] = [
+    '#type' => 'fieldset',
+    '#title' => t('Content Status Display'),
```
msayoung, line 101: I'm not sure about this heading , but I can't think of a better one! :)

localgov_base.theme
markconroy: Updated, good suggestion.

localgov_base.theme
markconroy: It's just a string. Let's leave it for now and we can easily change it later if we think of a better one.

markconroy:
@msayoung @tonypaulbarker This is ready for another review.

ctorgalson (approved):
:+1:

PR #830, opened 2025-07-17
Update README.md with notes on supported branches.

finnlewis:
Update README reflecting agreed support of branches.


tonypaulbarker:
@finnlewis I also made https://github.com/localgovdrupal/localgov_base/pull/829 and https://github.com/localgovdrupal/localgov_base/pull/828 

We need it in the 1.x readme too.

I didn't include the notes on how to upgrade grid because there may be other changes needed to be made between now and December

tonypaulbarker (approved):

PR #836, opened 2025-07-31
Remove show_title property to work around phpunit test failures in lo…

finnlewis:


finnlewis:
@markconroy @tonypaulbarker setting the default values before calling the include seems to solve the issue and allows the phpunit tests to run in localgov_microsites_group.

templates/block/localgov-blogs-prev-next-block.html.twig
```diff
@@ -1,2 +1,10 @@
 {% set prev_next_type = 'blog' %}
+{% set icon_path = icon_path|default('@localgov_base/includes/icons') %}
+{% set prev_icon = prev_icon|default('chevron-left') %}
+{% set next_icon = next_icon|default('chevron-right') %}
+{% set prev_label = prev_label|default('Previous'|t) %}
+{% set next_label = next_label|default('Next'|t) %}
+{% set prev_aria_label = prev_title ? (prev_label ~ ': ' ~ prev_title) : prev_label %}
+{% set next_aria_label = next_title ? (next_label ~ ': ' ~ next_title) : next_label %}
+{% set show_title = show_title|default(false) %}
```
markconroy: I think we should only need the `|default` in the original template. This template that is calling the SDC should just be setting the values that it needs like this:

```twig
{% set icon_path = '@localgov_base/includes/icons') %}
{% set prev_icon = 'chevron-left' %}
{% set next_icon = 'chevron-right' %}
{% set prev_label = 'Previous'|t %}
{% set next_label = 'Next'|t %}
{% set prev_aria_label = prev_title ? (prev_label ~ ': ' ~ prev_title) : prev_label %}
{% set next_aria_label = next_title ? (next_label ~ ': ' ~ next_title) : next_label %}
{% set show_title = false %}
```

I haven't tested this, but I think I am right.

markconroy (dismissed):

templates/block/localgov-blogs-prev-next-block.html.twig
tonypaulbarker: That's correct @markconroy - the defaults are for the SDC in case a template includes or embeds without specifying a value. 

`{% include 'localgov_base:prev-next' %} `

is not the correct way to include an SDC - I believe this syntax only includes the twig template like a regular include would.

It should instead be along the lines of this syntax to get the features of an SDC

{{ include('provider:my_component', {my_prop: my_prop_value}, with_context = false) }}

Perhaps this accounts for the funky behaviour we observed. Perhaps CSS isn't being read out of the box for 1.x but no-one noticed. In 2.x, I'm not sure this is a problem because we have 'embed' instead. 

We should review our includes and embeds.

tonypaulbarker:
Updated this for a re-test.

Workings: 

localgov_blogs.module sets these to NULL

Pushed a change that sets show_title to false in the template. 

I could not find some lovely secret sauce to handle the NULL value in the component but this bloated snippet did work:

`{% set show_title = show_title is not empty and show_title is not same as('') and show_title is not same as(false) and show_title is not same as('0') %}`

Follow ups:

localgov_blog

- We might follow up to change the module code to also set false instead of NULL. However, it might be that Drupal doesn't set a boolean for this variable. If Drupal provides the variable we should check the data type. The data type should match the SDC.
- The effect of NULL values (no previous or next) appears to be not to show previous next component. The block is not installed by default. We might consider deprecating and then removing this block feature from localgov_blog. 

localgov_base 2.x

- Maybe the variable should be a string and not a boolean. If Drupal provides the show_title variable for the block we should check the data type. The data type should match the localgov_blog module code.
- variable names prev_url (single directory component) and previous_url (module) do not match.
- We should check / update 2.x for this variable issue in this block
- We should review SIngle Directory Components includes and embeds
- We should improve checks and defaults in the SDCs
- We should probably use double curly brackets when we include SDCs

finnlewis:
@tonypaulbarker @markconroy my latest commit fixes the tests in localgov_microsites_group for me locally.

https://github.com/localgovdrupal/localgov_base/pull/836/commits/8319122baaf0ee3692afcc38edb44407e2676036

Does that look OK?

tonypaulbarker:
@finnlewis did the previous not work? 
The new commit will work too I think, but I was following recommended practice by using the function instead as include function is recommended over include tag.

See here:

https://www.drupal.org/docs/develop/theming-drupal/using-single-directory-components/using-your-new-single-directory-component

Example of how to include the sdc_examples:my-button component from a Drupal Twig template:

{{ include('sdc_examples:my-button', { text: 'Click Me', iconType: 'external' }, with_context = false) }}
You’ll need to provide the component ID ([module/theme name]:[component]) of the component.

with_context = false parameter is optional but recommended to avoid unexpected side-effects.

About {% include %} tag
**This tag is not recommended according to Twig documentation:**

It is recommended to use the include function instead as it provides the same features with a bit more flexibility:

The include function is semantically more "correct" (including a template outputs its rendered contents in the current scope; a tag should not display anything)
The include function is more "composable"
The include function does not impose any specific order for arguments thanks to named arguments.
Source: https://twig.symfony.com/doc/3.x/tags/include.html

finnlewis:
Thanks @tonypaulbarker , interesting!

No, the previous commit complained about syntax, which might just have been the `=` vs `:` in the assignment of the variables.

I'll flip the rest back and test that now.

finnlewis:
@tonypaulbarker @markconroy looking better? 

Note: the tests pass locally too with: 

```
phpunit web/modules/contrib/localgov_microsites_group/tests/src/Functional/GroupContentTypeAccessTest.php --filter=testContentContentTypeAccess
```

finnlewis:
Test pass here too, we good to merge this and tag a 1.x release?

Interesting to learn a bit more about best practice twig includes!

tonypaulbarker (approved):
@finnlewis this is ready

markconroy:
I think we should be targetting 2.x with thsi PR, not 1.x

finnlewis:
@markconroy the tests I'm trying to fix are from the 1.x branch. 

Because https://github.com/localgovdrupal/localgov_microsites_base/blob/2.0.9/composer.json#L9 

So I'd like to get this merged and released on the 1.x branch. 

The template is also quite different on the 2.x branch https://github.com/localgovdrupal/localgov_base/blob/2.x/templates/block/localgov-blogs-prev-next-block.html.twig

tonypaulbarker:
@markconroy @finnlewis We need a separate PR for 2.x the code for it will be slightly different to this. 

`{% set show_title = false %}`

tonypaulbarker:
@markconroy @finnlewis I opened a PR for 2x here https://github.com/localgovdrupal/localgov_base/pull/840

PR #837, opened 2025-08-01
fix: failed tests

millnut:
## What does this change?

This fixes the following issues https://github.com/localgovdrupal/localgov_publications/issues/255 and https://github.com/localgovdrupal/localgov_step_by_step/issues/130 as well as guides throwing the below error;

```
Drupal\Core\Render\Component\Exception\InvalidComponentException: [prevNextAttributes] String value found, but an object is required in Drupal\Core\Theme\Component\ComponentValidator->validateProps() (line 206 of core/lib/Drupal/Core/Theme/Component/ComponentValidator.php).

``` 

Also updated schema to the example recommendation https://www.drupal.org/docs/develop/theming-drupal/using-single-directory-components/annotated-example-componentyml#s-full-example

Can be tested by running locally with the profile and demo content for publications, step by step pages and guides

### DIFF
--- a/components/add-to-calendar/add-to-calendar.component.yml
+++ b/components/add-to-calendar/add-to-calendar.component.yml
@@ -1,5 +1,5 @@
 ---
-$schema: https://git.drupalcode.org/project/drupal/-/raw/10.1.x/core/modules/sdc/src/metadata.schema.json
+$schema: https://git.drupalcode.org/project/drupal/-/raw/HEAD/core/assets/schemas/v1/metadata.schema.json
 name: Add to Calendar
 description: Provides common add-to-calendar button with accompanying modal.
 status: stable
--- a/components/content/displays/display-full/display-full.component.yml
+++ b/components/content/displays/display-full/display-full.component.yml
@@ -1,2 +1,2 @@
 ---
-$schema: https://git.drupalcode.org/project/drupal/-/raw/10.1.x/core/modules/sdc/src/metadata.schema.json
+$schema: https://git.drupalcode.org/project/drupal/-/raw/HEAD/core/assets/schemas/v1/metadata.schema.json
--- a/components/icon/icon.component.yml
+++ b/components/icon/icon.component.yml
@@ -1,5 +1,5 @@
 ---
-$schema: https://git.drupalcode.org/project/drupal/-/raw/10.1.x/core/modules/sdc/src/metadata.schema.json
+$schema: https://git.drupalcode.org/project/drupal/-/raw/HEAD/core/assets/schemas/v1/metadata.schema.json
 name: Previous-next
 description: Provides common Previous-Next navigation element for LGD themes and modules.
 status: stable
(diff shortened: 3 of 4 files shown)


tonypaulbarker:
@millnut 

Related to https://github.com/localgovdrupal/localgov_base/pull/836 

Schema updates look good.

I haven't seen something this used in an SDC before:
`type: 'Drupal\Core\Template\Attribute'`

I did not try the cases suggested yet but plugged it into a microsite I had open and saw 

Drupal\Core\Render\Component\Exception\InvalidComponentException: [props.type] Does not have a value in the enumeration ["array","boolean","integer","null","number","object","string"]/n[props.type] String value found, but an array is required/n[props.type] Failed to match at least one schema in Drupal\Core\Theme\Component\ComponentValidator->validateDefinition() (line 122 of core/lib/Drupal/Core/Theme/Component/ComponentValidator.php).

so maybe only ["array","boolean","integer","null","number","object","string"] are valid types of prop.

markconroy:
I haven't seen `type: 'Drupal\Core\Template\Attribute'` used before either. 

If `object` is wrong, maybe we should use `string` for the prop type?

Might be worth a chat during merge tuesday today. Or maybe we create two PRs, one to update the schema and another to fix the prop type.

markconroy:
Update, I've just seen this in Olivero theme:

```
# Schema for the props. We support www.json-schema.org. Learn more about the
# syntax there.
props:
  type: object
  properties:
    attributes:
      type: Drupal\Core\Template\Attribute
      title: Attributes
      description: Wrapper attributes.
```

And that seems to be what is in this PR.

tonypaulbarker:
Right you are @markconroy - more testing needed.

Should the single quotes in `type: 'Drupal\Core\Template\Attribute'` be removed? 

Is there anything special about 'attributes' ? 

`attributes:` vs `prevNextAttributes:`

markconroy:
Olivero does not have single quotes. Do you want to test with those removed since you have microsites set up? Then we can potentially get this merged at Merge Tuesday.

tonypaulbarker:
I don't think the testing in my environment was reliable anyway. I am re-testing.

finnlewis:
Testing this now on microsites to see if it fixes the failing tests locally.

finnlewis:
This is against the 2.x branch and localgov_microsites_group is using localgov_microsites_base, which is still using localgov_base 1.x. https://github.com/localgovdrupal/localgov_microsites_base/blob/2.x/composer.json#L9

finnlewis:
Locally, testing this branch with localgov_microsites_group.
```
phpunit web/modules/contrib/localgov_microsites_group/tests/src/Functional/GroupContentTypeAccessTest.php --filter=testContentContentTypeAccess

```
Still gives me: 

```
Drupal\Core\Render\Component\Exception\InvalidComponentException: [show_title] NULL value found, but a boolean or an object is required. This may be because the property is empty instead of having data present. If possible fix the source data, use the |default() twig filter, or update the schema to allow multiple types. in Drupal\Core\Theme\Component\ComponentValidator->validateProps() (line 206 of core/lib/Drupal/Core/Theme/Component/ComponentValidator.php).
Drupal\Core\Template\ComponentsTwigExtension->doValidateProps(Array, 'localgov_base:prev-next') (Line: 110)
Drupal\Core\Template\ComponentsTwigExtension->validateProps(Array, 'localgov_base:prev-next') (Line: 48)
```

tonypaulbarker:
@finnlewis this doesn't affect show_title

I plugged 2.x into microsites. Confirmed it does not blow up there.

finnlewis:
Merge Tuesday, we're happy to merge this, thanks @millnut

PR #843, opened 2025-08-12
Adds aria label for directory facets blocks

### LINKED ISSUE #842: Add aria-label to directory search/facets if label is hidden
Adjacent to https://github.com/localgovdrupal/localgov_microsites_base/pull/324

If we merge this one, then we can delete the other one.

markconroy:
Closes #842

## What does this change?

Checks if the directory channel search block and/or directory channel facets block have their label hidden. If so, add the label as an aria-label so screen reader users will understand what these blocks are, since they come before the H1 in the page.

---

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -385,6 +385,17 @@ function localgov_base_preprocess_block(&$variables): void {
     }
   }
 
+  // Add a decent aria label to the facets blocks.
+  if (in_array($variables['base_plugin_id'], ['localgov_directories_channel_search_block', 'facet_block'])) {
+    if ($variables['elements']['#configuration']['label_display'] === '0') {
+      $routeMatch = \Drupal::routeMatch();
+      $node = $routeMatch->getParameter('node');
+      $node_title = $node instanceof NodeInterface ? $node->getTitle() : '';
+      $variables['attributes']['aria-label'] = t('Search') . ' ' . $node_title;
+      $variables['attributes']['role'] = 'search';
+    }
+  }
+
   // Add vertical or horizontal class to the guides navigation block.
   // In config, this will default to vertical for new installations, but remain
   // horizontal for existing installations.
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -386,7 +386,7 @@ function localgov_base_preprocess_block(&$variables): void {
   }
 
   // Add a decent aria label to the facets blocks.
-  if (in_array($variables['base_plugin_id'], ['localgov_directories_channel_search_block', 'facet_block'])) {
+  if (in_array($variables['base_plugin_id'], ['localgov_directories_channel_search_block', 'facet_block'], TRUE)) {
     if ($variables['elements']['#configuration']['label_display'] === '0') {
       $routeMatch = \Drupal::routeMatch();
       $node = $routeMatch->getParameter('node');


finnlewis:
@msayoung keen to test this and see how it applies to other facets like news.

finnlewis:
@tonypaulbarker  and @BHCC-hannahharvey also happy to take a look.

tonypaulbarker:
How does one go about toggling the labels please? 

I'm not sure the 'Directory facets' or 'Directory channel search' block labels are useful. I couldn't see how to toggle 'Search' or 'Country'

Thanks! 

Screenshot 2025-10-07 at 11 23 13

Screenshot 2025-10-07 at 11 22 51

markconroy:
@tonypaulbarker I _think_ those labels are part of the template and not toggleable.

If we want them toggleable, let's create a new issue, I'd like to get this aria one merged as soon as we can.

andybroomfield:
The facet group labels are part of the template. With Directory (and later Finder) facets, these are part of editor defined bundles, the facets themselves are actually just one list which the template splits up with the bundle header for each group.

Looking at our site, since we moved to Ajax facets those sit within the exposed form, but it's not producing a label (hidden or otherwise) for the filter or facets as a whole. Perhaps that's an oversight on Drupal side although each section does have a label.

tonypaulbarker:
@markconroy @andybroomfield please would you have another attempt at helping me understand a few things, I'm confused by this one.

The MR description says:
"Checks if the directory channel search block and/or directory channel facets block have their label hidden." 

By toggle, I mean, "How can I test by toggling the block label state from hidden to not hidden and vice versa?".

What do we mean by block label? Something other than the show title toggle on the block UI? Facets label? Where in the UI or otherwise can we toggle it? If in a template, why does the code appear to check configuration? 

Some step by step testing instructions would be great - so I know I am looking in the right place, i.e. what demo URLs should I be checking?

Thank you :)

finnlewis:
@msayoung might you be able to chime in here?

andybroomfield:
As this is the blocks themselves, not the individual categories, I think this PR is correct. Though I am curious why aria-label is not set or a .visually-hidden label is not being set, as I thought that was default behaviour.

andybroomfield:
Tested and I see the labelled blocks. Bit confusing as it duplicated the label for both blocks, when ideally these should be wrapped. But that's a consequence of the facets being a separate block.

Screenshot 2025-10-28 at 4 11 06 pm

Should the labels be that of the block configuration instead? And should they be different to each other. I also think a `<h2>` or `<h3>` that is visually hidden would be a better approach? I know aria-label can be frowned upon.

andybroomfield:
@tonypaulbarker As per merge Tuesday call, the labels such as country aren't toggled by a site builder as they are part of the template. The only label that can be toggled are the block headers "directory facets" or what a user puts into the block config form. This PR adds a separate label for the div for the block as they are hidden by default.

msayoung:
@markconroy - just giving this a spin again , I have some questions

- Should this affect only Directories or also other facets (News, Events etc) - currently it only seems to work on Directories. 

As @andybroomfield notes, where there are multiple blocks on the page (eg a search block and a facets block) they get the same aria-label.  I don't think this is ideal. I think we're going to have a further issue that on News we'll have multiple facet blocks which will all be named the same. 

This is partly a problem with different approaches to placing facets blocks, which we should probably _also_ address... (or maybe it's different / solved in finders?) 

> Should the labels be that of the block configuration instead? 

We chose not to go with the block config label in the other thread https://github.com/localgovdrupal/localgov_microsites_base/issues/322 - because the default (Directory Channel search / Directory facets) is meaningless to the end user. However, I'm think that perhaps that is a more sensible approach and we should rename the blocks, or find another way to encourage site builders to do so.  

Perhaps this needs talking through in real-time.

markconroy:
@msayoung Let's try set aside an hour next week to go through a real time walkthrough of it and see can get to a solution. Would be nice to have it checked off. 

Thanks.

markconroy:
https://www.drupal.org/project/localgov_base/issues/3563417

PR #846, opened 2025-08-12
Removes CSS that is not being used

### LINKED ISSUE #844: Remove superfluous css:
If the only content of `.link-wrapper` is `.link-content`, I think line 3 is superfluous, and line 4 belongs after line 8.

_Originally posted by @ctorgalson in https://github.com/localgovdrupal/localgov_base/pull/790#discussion_r2206005886_

markconroy:
Closes #844

### DIFF
--- a/css/components/link.css
+++ b/css/components/link.css
@@ -1,6 +1,5 @@
 .link-wrapper {
   position: relative;
-  display: flex;
   align-items: center;
 }
 
--- a/css/components/link.css
+++ b/css/components/link.css
@@ -1,6 +1,5 @@
 .link-wrapper {
   position: relative;
-  align-items: center;
 }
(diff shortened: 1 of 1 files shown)


css/components/link.css
```diff
@@ -1,6 +1,5 @@
 .link-wrapper {
   position: relative;
-  display: flex;
   align-items: center;
```
tonypaulbarker: I think that align-items: center without flex is redundant? The original issue has "If the only content of .link-wrapper is .link-content, I think line 3 is superfluous, and line 4 belongs after line 8." i.e move align-items: center to  '.link-content > a'

tonypaulbarker (changes requested):

PR #847, opened 2025-08-13
Feature/2.x/785 padding horizontal causes unwanted indenting

### LINKED ISSUE #785: Our .padding-horizontal class causes unwanted indenting
We have a very handy `.padding-horizontal` class so text/content does not bump up against the side of the screen on small screens.

However, since it's added at the page.html.twig level, and node.html.twig level, and paragraph.html.twig level, it can lead to ugly indenting which you need to override with CSS or custom templates.

Let's figure out a way to fix this, if possible.

ctorgalson:
# DRAFT

Closes https://github.com/localgovdrupal/localgov_base/issues/785

## What does this change?

This is a proof-of-concept PR demonstrating a way we can eliminate the hard-to-predict and hard-to-manage layout inconsistencies brought about by the use of `.padding-horizontal`.

---

It's finished enough for demonstration, but not ready for use.

---

The main addition to the theme is [a new SDC, `localgov_base:section`](https://github.com/localgovdrupal/localgov_base/pull/847/files#diff-8ce6cea4b7c9b0338c4aafd5c1b1340bb0f8504a88e95660853f6e259de3324b). The new SDC always renders the same markup, and can be configured to output [one or more of ten layout classes](https://github.com/localgovdrupal/localgov_base/pull/847/files#diff-ce1e26b211bf69024494948bb12327bcd4b899c6102610af56df871f8c3ddd08) that address most common layout issues.

For demonstration purposes, the branch makes significant changes to several blocks, the main [`page.html.twig` template](https://github.com/localgovdrupal/localgov_base/pull/847/files#diff-a7415b16ac9f95e74ef6652cf8bf086960a451b60618d22b7a8c6aaf5a6bc1dc), the [`region.html.twig` template](https://github.com/localgovdrupal/localgov_base/pull/847/files#diff-062f3a5548943152eedeef11e9add8372f763e39ffac8cd269d686bd0fca76c0), the `header.html.twig` template, and the `full-content` SDC, altering all of them to use the new SDC.

Even in the early stage the branch is in, most layout issues are gone. One notable exception to this is the `localgov_subsite_*` content types. The Page Builder paragraphs haven't been touched, though it wouldn't be a lot of further effort to get them to behave the same way.

I assume, but haven't checked in any detail, that the Scarfolk Council theme would need some/all of these changes as well.

## How to test

- checkout this branch of `localgov_base` in an LGD site
- set `localgov_base` as the default theme
- try to discover places (at any breakpoint) where there are inconsistencies in the amount of space between the left and right edges of the viewport and the content
- compare to the same URLs on an unmodified version of the site like https://demo.localgovdrupal.org/

Reduction or elimination of width inconsistencies in content.

The main risk is the reason why this change should only accompany a major version change to the theme: existing child themes _will_ have corrected some the issues this change corrects, but in different, incompatible ways.


components/section/section.css
```diff
@@ -0,0 +1,80 @@
+.lgd-section {
+  padding-inline: var(--spacing, 1rem);
+}
+
+.lgd-section__inner {
+  margin: 0 auto;
```
markconroy, line 6: margin auto here is causing all the section__inner items to be center aligned, this is affecting the main content block on service pages

components/section/section.css
```diff
@@ -0,0 +1,80 @@
...
+  max-width: var(--width-container, 73.75rem);
```
markconroy, line 7: I'm not sure exactly why, but this max-width is making our section__inn 1180px by default (e.g. breadcrumbs) but in the current theme (e.g. demo.localgovdrupal.org) they are equating to 1148px. 

It's probably because the padding-inline is now on the outer container instead of the inner container.

components/section/section.css
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/* BLEED: outer element has no inline padding */
+.lgd-section--bleed {
+  padding-inline: 0;
+}
+
+/* WRAPPED: inner element has block and inner padding */
+.lgd-section--wrapped > .lgd-section__inner {
+  padding-block: var(--vertical-rhythm-spacing, 1.5rem);
```
markconroy: I wonder should we create a custom variable for this, something like `--section-block-spacing` and set that to equate to `--vertical-rhythm-spacing` in case some designs wants extra/less padding here, but still want the vertical rhythm amount for paragraphs and things.

templates/block/block--local-tasks-block.html.twig
```diff
@@ -7,8 +7,17 @@
...
 {% block content %}
   {% if content %}
-    <nav class="tabs" role="navigation" aria-label="{{ 'Tabs'|t }}">
-      {{ content }}
-    </nav>
+    {% embed 'localgov_base:section' with {
+      section_inner_classes: [ 'tabs' ],
+      section_outer_attributes: create_attribute({
+        'aria-label': 'Tabs'|t,
+        role: 'navigation',
```
markconroy: I think we can leave out `role="navigation"` since it's implied by the `nav` element.

templates/block/localgov-alert-banner.html.twig
```diff
@@ -69,5 +70,6 @@
         </div>
       {% endif %}
     </div> {# End Inner #}
-  </div> {# End Wrapper #}
-</article>
\ No newline at end of file
+
+  {% endblock %}
+{% endembed %}
```
markconroy, line 75: We'll need a blank line here at the end of file so the tests don't fail

markconroy (changes requested):
Thanks @ctorgalson, this looks like a great piece of work.

I've left a few small changes that we should look at and then do some more testing. 

Would be great to get this and a few more things into a new major release (e.g. remove the components directory that is now using SDC instead)

components/section/section.css
ctorgalson: This one should be fixed on the current branch.

components/section/section.css
ctorgalson: This is true for _every_ region/thing that's not restricted width (see screenshot). I based it on the header which _is_ 1180px width in the current version of the theme.

In any case, we'd just need to set a _single_ variable to change it to 1148 or any other value.

image

components/section/section.css
ctorgalson: Do we need an extra variable? This is already achievable with a single line of css in the theme's `variables.css` or elsewhere:

```css
/**
 * Using my new favourite unit for this, `lh`.
 *
 * @see https://caniuse.com/mdn-css_types_length_lh
 */
.lgd-section {
  --vertical-rhythm-spacing: 1lh;
}
```

templates/block/block--local-tasks-block.html.twig
ctorgalson: Fixed in current branch.

templates/block/localgov-alert-banner.html.twig
ctorgalson: Fixed in current branch.

ctorgalson:
@markconroy, @tonypaulbarker what do we think of the overall approach?

My thoughts:

**Pros**

- completely capable of solving the specific issue (including Page Builder components, though I haven't done that work in this PR)
- css-wise, quite simple and easily-documented
- _reasonably_ intuitive (though see "naming" under "Cons")
- handles both page-level and content-level layouts more easily than I'd anticipated
- plays nice with grid formatting (i.e. since it's trivial to make `.lgd-section`/`.lgd-section__inner` elements without padding)

**Cons**

- naming is pretty wordy, and not necessarily descriptive enough
- element class-bloat
- a _little_ markup bloat (i.e. since every place it's used it inserts two `<div>` elements)
- [NIH](https://en.wikipedia.org/wiki/Not_invented_here) syndrome :grimacing:? (I don't really like e.g. Tailwind, but I'm sure it [or something of that sort](https://hackr.io/blog/best-css-frameworks) could handle this job, and we're proposing breaking changes anyway...)
- I'm still not completely convinced that utility classes are the best way to solve the general problem(see below)

```css
/**
 * Solving the basic issue with (almost) no utility classes 
 *
 * Given a pre-established set of layouts, we can create the utility classes (like e.g. `region--full-width`), but then
 * add selectors to that class's css rule.
 *
 * By doing this, we can essentially eliminate most uses of utility classes. And since localgov_base acts more like
 * a starterkit theme than a base theme, we can always have a local copy of this stylesheet.
 */

/**
 * Full width regions
 *
 * Use .region--full-width utility class, OR add selectors to this list.
 */
.region--full-width,
.region--header,
.region--footer,
.region--content {
  padding-inline: var(--spacing);
}

/**
 * Constrained content regions
 *
 * Use .region--constrained utility class, OR add selectors to this list.
 */
.region--constrained,
.region--content {
  margin-inline: auto;
  max-width: var(--width-container);
   
  /* Variations by node bundle */
  .node--type--localgov-services-page & {
    margin-inline: 0 calc((100vw - var(--width-container)) / 2 + (var(--width-container) - var(--width-restricted-width-section))));
    max-width: var(--width-restricted-width-section);
  }
}
```

ctorgalson:
@markconroy, @tonypaulbarker any thoughts?

markconroy:
@ctorgalson are you coming to tech group drop in today? We could chat about it there.

ctorgalson:
@markconroy I had another call. I'll come on the 9th for sure.

ctorgalson:
Sorry @markconroy, last-minute call here.

PR #849, opened 2025-08-25
Uses document/media name instead of file name if available

### LINKED ISSUE #848: Use document name if available instead of file name

markconroy:
Closes #848 

## What does this change?

If the name field has been filled in for a document media item, it will use this instead of the file name when displaying files, e.g. "My File (pdf, 2MB)" instead of "my-file.pdf (pdf, 2MB)"

---


tonypaulbarker (approved):
Works well. Why isn't this in core?!

rupertj:
@tonypaulbarker This is in core. There's an option on the media display mode for it.

Edit: Ok, the option there is to use the description field, not the media title, but it is a simple little setting that effectively does the same thing.

PR #857, opened 2025-09-15
Adds basic CSS variables settings for theme

### LINKED ISSUE #57: Allow a sub section of our variables to be set via the theme UI
I think it might be good to allow editors to set some variables via the theme - for simple one-off changes that might last a day or two, e.g. accent colour (you might want to change it to red for christmas!), fonts (if we provide a set of prepared fonts like in microsites), etc.

This will also give us a good base to try bring live previews module to LGD core.

markconroy:
Closes localgovdrupal/localgov_scarfolk#57 

## What does this change?

Adds theme settings for some CSS variables:

- colour accent
- line-height
- base spacing

## How to test

Open the settings page of your theme, set some values in the "Theme settings" fieldset. Go back and check they have taken effect on the frontend.

---

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -145,6 +145,35 @@ function localgov_base_form_system_theme_settings_alter(&$form, FormStateInterfa
     '#description'   => t('This will display the guide navigation vertically above the guide.'),
   ];
 
+  $form['localgov_base_settings']['theme_styles_fieldset'] = [
+    '#type' => 'fieldset',
+    '#title' => t('Theme Styles'),
+    '#description' => t('Control various theme styles.'),
+  ];
+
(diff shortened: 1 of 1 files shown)


tonypaulbarker:
Today on the product drop in we discussed where these settings should be and the idea of adding these settings to scarfolk theme rather than localgov_base. 

If the intention is for early prototyping only, using Scarfolk instead could prevent mishaps making it into production (unless people are using Scarfolk in production of course). 
If the intention is for production use perhaps they stay here. 
If we will make a prototyping add on for other cases like arbitrary CSS, perhaps the settings should live there instead.

finnlewis:
Marking this as draft , moving to Scarfolk

markconroy:
Closing in favour of https://github.com/localgovdrupal/localgov_scarfolk/pull/58

PR #859, opened 2025-09-15
Adds theme setting for extra CSS

### LINKED ISSUE #858: Allow extra CSS be added via the UI
It _might_ be a good idea for people to be able to add some arbitrary CSS to the theme via the UI - e.g. early days, before you have a developer, but want the theme to have a bit of council branding.

I understand we probably don't want people doing this for production sites, but having the facility during dev could be welcome.

markconroy:
Closes #858 

## What does this change?

Allows people to add extra CSS to the site via the theme settings.

---

### DIFF
--- a/config/schema/localgov_base.schema.yml
+++ b/config/schema/localgov_base.schema.yml
@@ -41,3 +41,8 @@ localgov_base.settings:
       label: 'Responsive WYSIWYG tables'
       description: 'Choose whether to make WYSIWYG tables responsive or not.'
       default_value: TRUE
+    localgov_base_extra_css:
+      type: string
+      label: 'Extra CSS'
+      description: 'Add any additional CSS styles you want to include.'
+      default_value: ''
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -145,6 +145,19 @@ function localgov_base_form_system_theme_settings_alter(&$form, FormStateInterfa
     '#description'   => t('This will display the guide navigation vertically above the guide.'),
   ];
 
+  $form['localgov_base_settings']['theme_styles_fieldset'] = [
+    '#type' => 'fieldset',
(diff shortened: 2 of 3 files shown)


rupertj:
The problem with this as-is it that there's no filtering or validation of what's put in, for example, you could put this in:
```
</style>
<script src="https://example.com/cross_site_scripting_attack.js"></script>
<style>
```
I think if there's a desire to have this, it should validate and sanitise the CSS correctly.

tonypaulbarker:
Not keen on this approach - even without potential security holes, maintaining sites with arbitrary code injected via CSS injector and asset injector contrib modules can be something of a challenge.

The discussion on slack is for a use case for making quick prototyping easier and consensus seems to be that this should never be used in production. If it's included in localgov_base it inevitably will be. 

I suggest closing this MR and gathering requirements / solutionising that doesn't compromise the end product. The feature request seems to be "how can we do rapid prototyping without needing acces to the code / dev tools?".

finnlewis:
Discussing at Merge Tuesday: 
 - Mark would not recommend this on a live production site.
 - Tony suggesting that this might be better as a separate module, to allow for prototyping / testing.
 - Maria: 2 issues 1. prototyping 2. something minor for subsites... hex codes for certain colours for example. More boradly, can we review the concept holistically, where might site managers want to manage certain css / visual settings theough the browser?
 - Andy: module css_editor https://www.drupal.org/project/css_editor - could be a route for allowing folks to do this. 
 - @willguv - what are the design tinkering needs on a production site?

finnlewis:
From discussion in the Product Drop-in yesterday, there was a general feeling that we should move these sort of point and click css changes / injecting CSS to the Scarfolk theme for now to avoid tempting people to do this on a production site.

PR #863, opened 2025-09-17
check media item exists before loading it

### LINKED ISSUE #862: InvalidArgumentException: Field mid is unknown.
InvalidArgumentException: Field mid is unknown. in Drupal\Core\Entity\ContentEntityBase->getTranslatedField() (line 616 of core/lib/Drupal/Core/Entity/ContentEntityBase.php). 

Caused again by `localgov_base_preprocess_file_link()`

I installed the `localgov_irish_planning_notices` module, added a file to a notice, and got the above issue.

This seems to fix it:

```php
  // Load media entity from file.
  $entity = $file->_referringItem?->getParent()?->getParent()?->getEntity();
  $media = ($entity instanceof MediaInterface) ? $entity : NULL;

  if (isset($variables['link']['#title'])) {
    $document_title = $variables['link']['#title'];

    if ($media) {
      $media_name = $media->get('name')->getString();
      // If the media name is set and different to the
      // $variables['link']['#title'], we will use that instead.
      $document_title = ($media_name && $media_name !== $document_title) ? $media_name : $document_title;
    }

    $variables['link']['#title'] = [
      '#markup' => "{$document_title} <span class=\"file-meta\">(<span class=\"file-type\">{$variables['file_type']}</span>, <span class=\"file-size\">{$variables['file_size']}</span>)</span>",
    ];
  }
  else {
    $file_title = $variables['link'];
    $variables['link'] = [
      '#markup' => "{$file_title} <span class=\"file-meta\">(<span class=\"file-type\">{$variables['file_type']}</span>, <span class=\"file-size\">{$variables['file_size']}</span>)</span>",
    ];
  }
```

markconroy:
Closes #862 

## What does this change?

I removed the attempt to get the media ID using the non-existent mid field:

```php
$media_id = $file->_referringItem?->getParent()?->getParent()?->getEntity()?->get('mid')?->getString();
$media = $media_id ? Media::load($media_id) : NULL;
```

Instead, I directly check if the entity is a Media entity:

```php
$entity = $file->_referringItem?->getParent()?->getParent()?->getEntity();
$media = ($entity instanceof \Drupal\media\MediaInterface) ? $entity : NULL;
```

This change eliminates the need to get the media ID and then load the media entity again. It directly uses the entity if it's already a Media entity, which is more efficient and avoids the error with the non-existent mid field.

---

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -8,7 +8,6 @@
 use Drupal\Component\Utility\Crypt;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Url;
-use Drupal\media\Entity\Media;
 use Drupal\media\MediaInterface;
 use Drupal\node\NodeInterface;
 use Drupal\views\ViewExecutable;
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -423,13 +423,13 @@ function localgov_base_preprocess_file_link(&$variables): void {
   $variables['file_size'] = strtr($variables['file_size'], [' ' => '']);
 
   // Load media entity from file.
-  $media_id = $file->_referringItem?->getParent()?->getParent()?->getEntity()?->get('mid')?->getString();
-  $media = $media_id ? Media::load($media_id) : NULL;
+  $entity = $file->_referringItem?->getParent()?->getParent()?->getEntity();
+  $media = ($entity instanceof MediaInterface) ? $entity : NULL;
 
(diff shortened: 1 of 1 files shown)


tonypaulbarker (approved):
Confirm that I see the same issue and the error goes away with this fix.

Functional testing and re-testing:

- uploading planning notices
- file upload on a node form
- webform document upload

PR #872, opened 2025-10-20
only print link if it has a value

### LINKED ISSUE #866: Call out box link prints when the link field is empty.
The link is printing whether or not there is another in the link field. 

In paragraph--localgov-call-out-box.html.twig the localgov_button needs an if statement. 

The link field is not required. 

```
  <div class="call-out-box__content">

    {% if heading['#context']['text'] %}
      {{ heading }}
    {% endif %}

    {{ content|without('localgov_button', 'localgov_background_image', 'localgov_colour_theme', 'localgov_header_text', 'localgov_heading_level', 'localgov_opens_in_a_new_window') }}
    <a href="{{ button_url }}" class="call-out-box__link"
      {% if open_in_new_window %} target="_blank" {% endif %}>
      <span>{{ paragraph.localgov_button.getValue[0].title }}</span>
      {% if open_in_new_window %}
        <span class="visually-hidden"> {{ '(opens in a new window)'|t }}</span>
      {% endif %}
    </a>
  </div>
```

markconroy:
Closes #866 

## What does this change?

1. Adds an `if` clause to only print the link field if it has a value
2. Remove the space from below the body of the call out 
3. Add the space to the link field instead, so it's only available if there is an actual link

Call out - no link:

Screenshot 2025-10-20 at 11 03 44

Call out - with link:

Screenshot 2025-10-20 at 11 03 49

---

### DIFF
--- a/css/components/callout-box.css
+++ b/css/components/callout-box.css
@@ -18,7 +18,14 @@
   background: var(--call-out-box-bg-color);
 }
 
+.call-out-box__content .field--name-localgov-body-text > :last-child {
+  /* Ensure we don't have extra space if there is no call out link */
+  margin-block-end: 0;
+}
+
 .call-out-box__link {
+  display: inline-block;
+  margin-block-start: var(--vertical-rhythm-spacing);
   padding: var(--call-out-box-link-padding);
   color: var(--call-out-box-bg-color);
   background-color: var(--call-out-box-text-color);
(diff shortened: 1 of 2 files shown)


tonypaulbarker (approved):
Looks good to me. Before these changes we get an empty button styled link with no link in it that looks broken. After changes looks neat.

PR #876, opened 2025-10-20
Feature/875 update linters

### LINKED ISSUE #875: Our linters are no longer working
Because they are relying on Drupal's set up which has been updated.

markconroy:
Closes #875 

## What does this change?

Updates packages to match Drupal core's packages.

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,43 +14,116 @@
         "eslint-config-prettier": "^9.0.0",
         "eslint-formatter-gitlab": "^5.0.0",
         "eslint-plugin-import": "^2.25.4",
-        "eslint-plugin-jquery": "^1.5.1",
+        "eslint-plugin-no-jquery": "^3.0.2",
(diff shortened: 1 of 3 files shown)


finnlewis:
Just thinking out loud while looking at this.

Is this duplicating dependencies that Drupal core should cover? 

For example, we run the install frontend here: https://github.com/localgovdrupal/localgov_project/blob/3.x/.ddev/commands/web/install-frontend

Might we be able to remove the dependencies from this theme and update the documentation to follow that pattern.

Use Drupal core's linting config.

tonypaulbarker:
We discussed that ideally we would run the same tools from one place at the localgov_project level and remove these tools from individual projects. 

In the dicussion Mark mentioned that some teams are not using ddev. Can we support them in other ways (e.g. another method to run tools at the project root or some docs for adding linters to their sub themes).

PR #878, opened 2025-10-27
Prefer file description field for document link text if set and available

### LINKED ISSUE #871: Regression: Option to use the description for media links instead of the filename no longer works.
Image

The screenshot shows media embedded via the WYSIWYG. The links used to use the description of the media, but now use the filename, despite the option being checked to use the description.

I've seen the link get given the wrong title in localgov_base_preprocess_file_link(), which had recent changes. I'll investigate further to see they were the cause.

tonypaulbarker:
Closes https://github.com/localgovdrupal/localgov_base/issues/871

Continutes work to render the document files for the case where configuration provided in Drupal Core and known to be used on LGD sites prefers the description field to be used for the document link.

Requires code review.

Requires robust functional testing of the display in localgov_base of: 

Document media configured to use name field
Document media configured to use description field
Document media with no name or description field
Drupal file upload fields
Webform file uploads

### Intended logic 

Use description if " use_description_as_link_text " is set in document media manage display and description field has value.

Otherwise 

Use name field with fall backs as before

### DIFF
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -429,17 +429,17 @@ function localgov_base_preprocess_file_link(&$variables): void {
     $document_title = $variables['link']['#title'];
 
     if ($media) {
-      // Get the current view mode and field display settings.
-      $view_mode = $variables['element']['#view_mode'] ?? 'default';
-      $field_name = $variables['element']['#field_name'] ?? NULL;
+      // Assume some defaults for document media type.
+      $view_mode = 'default';
+      $media_file_field_name = ($media->bundle() === 'document') ? 'field_media_file' : NULL;
 
       // Initialise this variable as false.
       $use_description = FALSE;
 
-      if ($field_name) {
+      if ($media_file_field_name) {
         $entity_display_repository = \Drupal::service('entity_display.repository');
         $entity_display = $entity_display_repository->getViewDisplay($media->getEntityTypeId(), $media->bundle(), $view_mode);
-        $field_display = $entity_display->getComponent($field_name);
+        $field_display = $entity_display->getComponent($media_file_field_name);
 
         if ($field_display && isset($field_display['settings']['use_description_as_link_text'])) {
           $use_description = $field_display['settings']['use_description_as_link_text'];
--- a/localgov_base.theme
+++ b/localgov_base.theme
@@ -6,6 +6,7 @@
  */
 
 use Drupal\Component\Utility\Crypt;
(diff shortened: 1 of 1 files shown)


rupertj:
I've had a quick look at this - we don't need to get the description and set it as the title if the option's checked, as it's already been done by this point in the request.

If use_description_as_link_text is set, returning early and not doing anything else should be enough.

tonypaulbarker:
@rupertj thanks for having a look. use_description_as_link_text is probably true for almost every site since it has been in the default configuration, regardless of whether they have the description field or not to fill out. We do need to also see if the description is present. But maybe we can then return if it is.

tonypaulbarker:
Some weirdness in that we only modified localgov_base but CKEditor doesn't render the description for its label as expected when the field settings for the file description are active. The default media mode being a responsive image size doesn't make sense for files - should be 'default' instead.

rupertj:
I've tested this in H&F's site where the bug showed up and it's working well for me.

I've also exposed the media name field on it and disabled the option to use the file's description and it's showing the media name for media where I've set it.

There's a few spots where I think the code could be simpler and easier to follow, so I'll start a review.

In general though, I'd be happy for this to be merged as-is once the tests pass.

localgov_base.theme
```diff
@@ -429,10 +429,40 @@ function localgov_base_preprocess_file_link(&$variables): void {
...
+        $entity_display = $entity_display_repository->getViewDisplay($media->getEntityTypeId(), $media->bundle(), $view_mode);
+        $field_display = $entity_display->getComponent($media_file_field_name);
+
+        if ($field_display && isset($field_display['settings']['use_description_as_link_text'])) {
+          $use_description = $field_display['settings']['use_description_as_link_text'];
+        }
+      }
+      
+      // Get the description from the file field if available.
+      if ($use_description && $file->_referringItem) {
```
rupertj: This clause and the next one could be combined into one.

finnlewis:
> I've tested this in H&F's site where the bug showed up and it's working well for me.
> 
> I've also exposed the media name field on it and disabled the option to use the file's description and it's showing the media name for media where I've set it.
> 
> There's a few spots where I think the code could be simpler and easier to follow, so I'll start a review.
> 
> In general though, I'd be happy for this to be merged as-is once the tests pass.

Based on Ru's comment, let's merge this and refactor code later if needed! Hope that's OK @rupertj :)

PR #883, opened 2025-11-03
adds ckeditor stylesheet to subtheme generator

### LINKED ISSUE #882: Add a library for ckeditor styles in the subtheme generator
This could be helpful for others to know it exists in case they want to add some styles into their subtheme's ckeditor.

markconroy:
Closes #882 

---


NikLP:
Haven't tested the actual commit, but having employed the same changes in my local subtheme, I can report that the method is sound, so LGTM.

scripts/subtheme-items/css/base/ckeditor5.css
```diff
@@ -0,0 +1,17 @@
+/*
+  @file CKEditor styles.
+  These styles are applied within the CKEditor 5 interface to
```
tonypaulbarker, line 3: I think the guidance here should be to use sparingly, for cases where the editor could not otherwise tell how the content would be formatted (like style plug ins). The CKEditor window is not meant to match the front end (fonts, colours, spacing and so on) and making it match in some ways but not others can cause confusion. One example of a case is that heading colours may be different for different sections of the site. A suggestion would be to include a real example from the base theme file again here instead of the heading colours.

I think that linters might complain about a file with only comments?

scripts/subtheme-items/_subtheme.info.yml_
```diff
@@ -3,6 +3,13 @@ description: "Your sub-theme, which uses localgov_base as its base theme."
 type: theme
 core_version_requirement: ^10 || ^11
 base theme: "localgov_base"
+ckeditor5-stylesheets:
+  # List any stylesheets you want to apply to the CKEditor
+  # interface here. This allows you to match your content editing
```
tonypaulbarker, line 8: See review comments on scripts/subtheme-items/css/base/ckeditor5.css , applies here too.

tonypaulbarker (changes requested):
Left some thoughts on the code comment guidance for use cases and examples.

PR #887, opened 2025-11-10
adds options for 3 or 4 items in box links listing

### LINKED ISSUE #886: Allow options for 2 or 3 or 4 items in a row for box links
This is a sibling issue to https://github.com/localgovdrupal/localgov_paragraphs/issues/265

Both need to be merged at the same time.

markconroy:
Closes #886 

---

### DIFF
--- a/templates/field/field--paragraph--localgov-paragraphs--localgov-box-links-listing.html.twig
+++ b/templates/field/field--paragraph--localgov-paragraphs--localgov-box-links-listing.html.twig
@@ -42,11 +42,11 @@
 
 {% set classes = [
     'box-links-listing__list-item',
-  ] 
+  ]
 %}
 {% for item in items %}
-  {# 
-    We are checking the parent here, and then using whatever layout style is chosen 
+  {#
+    We are checking the parent here, and then using whatever layout style is chosen
     to set the width of the items.
   #}
   {% set parent = item.content['#paragraph']._referringItem.parent.parent.entity %}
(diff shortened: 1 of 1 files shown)


finnlewis:
Note: this will only have an effect we merge https://github.com/localgovdrupal/localgov_paragraphs/pull/267

@tonypaulbarker asks about update hooks, at the mo this has none.

Maybe just release notes to clarify.

finnlewis (approved):
Thanks @markconroy !
The layout options seem to work as expected. 

1 col

image

2 per row

image

3 per row

image

4 per row

image

finnlewis:
@markconroy I noted on the 3 per row example, that the spacing between the rows was a little different to the follow us links below. I think this is because they are using flex rather than grid, so the calculations are a little different, is that right?

markconroy:
@finnlewis it's because those box links are not a "box links" (plural) component. They are 3 individual box links in a 3-column layout. And our 3-column layouts are not using grid just yet.

tonypaulbarker:
@finnlewis @markconroy in the drag and drop editor of a subsite overview these don't have any spacing between them in the edit window so the settings don't appear to make any difference until its rendered on the front end. I think we should add some styles to localgov_admin_theme_improvements to handle the case because it might confuse editors. 

When I choose 'two per row' and I have three links it still displays three side by side in the editor, and I get some padding.

markconroy:
@tonypaulbarker if you want that spacing to appear on the backend, it might be best to add the CSS into the localgov_paragraphs module.

But then we'll have the question, "should we do this for all CSS?".

I think it's okay to have an 'admin experience' which focusses on content and as frontend experience which deals with design.

PR #889, opened 2025-11-11
Adds correct padding for media with text

### LINKED ISSUE #888: Media with Text (media positioned right) needs padding fixes
Our media with text component CSS is designed to work on large screens, but the lack of padding-inline-start when the media is on the right causes unwanted bucking up against the side of the component on small screens. Like this:

Image

It should look like this:

Image

PR Ahoy!

markconroy:
Closes #888 

---

### DIFF
--- a/css/components/media-with-text.css
+++ b/css/components/media-with-text.css
@@ -54,12 +54,8 @@
   width: 100%;
 }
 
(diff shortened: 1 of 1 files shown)


finnlewis (approved):
Before: 

image


After: 

image


Left padding more consistent :)

Thanks @markconroy

PR #892, opened 2025-11-12
check for value in blog body before printing div

### LINKED ISSUE #891: Don't print blog channel sidebar if the body is empty

markconroy:
Closes #891


tonypaulbarker:
This works as intended. Is there a case for the aside left in place to maintain layouts at all (looks fine in the demo)

finnlewis:
@tonypaulbarker is happy with this :)

PR #898, opened 2025-12-01
allowing configuring how many top tasks per row

### LINKED ISSUE #897: Have different numbers of top tasks and content columns on a Service landing page
In some councils the design of the top task buttons makes them look like column headings

See this [Croydon example](https://www.croydon.gov.uk/adult-health-and-social-care)

Image

We could consider:

- default settings of 4 top task buttons, 3 columns
- ability to councils to set what they like here
- leaving this alone. This [Cumberland example](https://www.cumberland.gov.uk/bins-recycling-and-street-cleaning) shows how to fix by including top tasks, page title and description on a background color

Image

### This is done when...

- We've decided what to do about this
- Fixes are written up
- Fixes are available for testing by selected users

### How this was requested

The following changes were discussed at the Content Group. See [Miro board](https://miro.com/app/board/uXjVMZSpkt4=/?moveToWidget=3458764602544478124&cot=14)

markconroy:
Closes #897

## What does this change?

Add a configuration option for how many top tasks we want to show per row for service landing pages, with the default being 4-in-a-row. 

Currently we have 3, so will need to have a note on the release notes that to keep the same feature we currently have, you will need to go to your theme settings and set it to `3` then save the form and export the config.

## How to test

- Go to theme settings page
- Scroll to bottom
- Set how many items per row for top tasks

---

### DIFF
--- a/config/install/localgov_base.settings.yml
+++ b/config/install/localgov_base.settings.yml
@@ -8,3 +8,4 @@ localgov_base_localgov_guides_stacked_heading: FALSE
 localgov_base_localgov_guides_vertical_navigation: TRUE
 localgov_base_mobile_breakpoint_js: '768'
 localgov_base_responsive_wysiwyg_tables: TRUE
+localgov_base_services_top_tasks_per_row: 4
--- a/config/schema/localgov_base.schema.yml
+++ b/config/schema/localgov_base.schema.yml
@@ -41,3 +41,8 @@ localgov_base.settings:
       label: 'Responsive WYSIWYG tables'
       description: 'Choose whether to make WYSIWYG tables responsive or not.'
       default_value: TRUE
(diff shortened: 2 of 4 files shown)


willguv:
We need the new sites to have 4 top tasks per row, but we don't want the existing sites to change. They need to opt in to new stuff

@markconroy proposed a helper module a while back for LG Base to apply updates - I think we need to seriously consider this as there will be many more changes like this in the future eg when we follow GOV.UK or NHS design system changes

@tonypaulbarker seconds this approach

markconroy:
PR to add updates module: #900 

We'll also need to have the module enabled.

markconroy:
https://www.drupal.org/project/localgov_base/issues/3563416

